PR #4, opened 2017-09-03
MAINT: obtain numpy-distutils from numpy main repository

pv:
The necessary distutils changes were merged to master, so we can get
them there.


matthew-brett:
Thanks - in it goes.

PR #5, opened 2017-09-14
Insert bundled library licenses to generated wheels

pv:
Append license texts to `LICENSE.txt`, for all libraries bundled in the wheels at the build stage.

I understood the list of bundled libraries is complete, but please correct if there's something missing --- I'm not familiar with all stuff what goes in the manylinux build.

ping: @rgommers


config.sh
```diff
@@ -66,3 +66,13 @@ function run_tests {
     # Show BLAS / LAPACK used
     python -c 'import scipy; scipy.show_config()'
 }
+
+function patch_code {
```
matthew-brett: Actually `config.sh` doesn't get sourced until after this.   I guess it could be sourced before by default.   But - for now - can you put this in it's own file?  As in:

```
./patch_code.sh $REPO_DIR
```

in the travis.yml file?

pv: ok, did it like that

matthew-brett:
I just checked in the built wheels over at https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com .  I am not completely sure which is which, but the wheels there currently don't seem to have the right OSX / Linux license text.  Worth checking with a test maybe?

pv:
PR builds cannot upload anything, because they don't have access to the upload keys.
The wheels on that site are not from this build.

matthew-brett:
Ah yes, you're right.  But nevertheless, it would be good to put a check in the tests, that the LICENSE file has the expected text, on the lines of:

```
SP_INSTALL_DIR=$(dirname $(python -c 'import scipy; print(scipy.__file__)'))
if [ "$(cat $SP_INSTALL_DIR/LICENSE.txt | grep OpenBLAS)" == "" ]; then
...
```

pv:
Added a check.

The Travis builds for this repository seem to be at the end of a really long queue. Do you know why this is? Is it affected by the presence of daily builds?

matthew-brett:
Thanks for that nice check.

Yes, there are lots of daily builds, which all happen to be around now.  I guess we could move this repo to the scipy organization, but then it would be competing with the scipy per-commit builds.

pv:
builds seem ok

matthew-brett:
Yes, looks good - thanks for doing that.

PR #8, opened 2017-09-18
Fix branch name, update for latest procedure

matthew-brett:


pv:
I'll merge this --- if there's more to change that can be done later.

PR #9, opened 2017-09-18
DOC: remove http links

pv:
Remove all http links from README.rst, especially wheels.scipy.org ---
we should follow good development practices.


matthew-brett:
I think most of these are in #8 ...

pv:
ok, sorry, I didn't check the existing prs...

matthew-brett:
I added a s to an http link I missed, from your PR.

PR #11, opened 2017-09-28
MAINT: pin versions of all build dependencies

pv:
Pin build dependencies to "known good" versions.

The argument for pinning build dependencies in this repository is:

The point is to be able to build wheels in a controlled way that we
can be sure works, such that e.g. Cython bugs don't torpedo 
the release process.

Testing whether there are incompatibilities in build dependencies (e.g.
whether latest Cython version works) on the other hand should be done
in the main CI repository.

@rgommers: ping --- I think it would be best to pin the deps also in the daily branch. This way, there is a continuous testing that the process that's going to be used for the releases produces usable wheels.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ env:
         - BUILD_COMMIT=v0.19.1
         - PLAT=x86_64
         - NP_BUILD_DEP="numpy==1.8.2"
+        - CYTHON_BUILD_DEP="Cython==0.26.1"
         - NP_TEST_DEP="numpy==1.8.2"
         - UNICODE_WIDTH=32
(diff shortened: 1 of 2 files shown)


rgommers:
> @rgommers: ping --- I think it would be best to pin the deps also in the daily branch. This way, there is a continuous testing that the process that's going to be used for the releases produces usable wheels.

Yes, on second thought I agree. We pin numpy, so should also pin cython. 

We do probably need to test against cython master in the main scipy CI, just as we test against numpy master.

pv:
ok, I'll merge this now (also stops travis from mailing me every day :)

pv:
I'll also put it in the master branch --- if we need to change the Cython version used there's still time and rc2 in the pipeline...

PR #12, opened 2017-10-04
Switch numpy-distutils to a version that bundles msvcp140.dll in the wheels

pv:
Bundle msvcp*.dll in the wheel to address https://github.com/scipy/scipy/issues/7969

numpy.distutils patch from https://github.com/numpy/numpy/pull/9819


pv:
oops, wrong branch

PR #13, opened 2017-10-04
Switch numpy-distutils to a version that bundles msvcp140.dll in the …

pv:
Bundle msvcp*.dll in the wheels to address https://github.com/scipy/scipy/issues/7969

The bundling is done manually just by copying the dlls to a place from which they'll be included in the wheel.

I didn't find redistributable msvcp100.dll for python 3.4 in the compiler distributions on appveyor, so those wheels aren't going to get the dll. Hopefully, few people use this Python version and we can just ignore this issue.

TODO:

- [x] Update LICENSE_win32.txt
- [ ] Check it actually works (!)


pv:
If this works, I'll just merge it asap.

PR #14, opened 2017-10-05
Merge daily to master

pv:
Onboard the `msvcp*.dll` bundling to master, if we are happy with it.
Probably needs to be done before rc2 in any case.

(Note that this PR builds fail because master is at 1.0.0rc1 which doesn't have the mpmath 1.0.0 test fix.)


pv:
Merging to be in time for rc2

PR #18, opened 2017-10-13
Update comment in `build_osx_wheel`.

dhermes:
Bash function no longer builds 32-bit wheels on OS X, but the comment indicated otherwise.


matthew-brett:
Thanks for spotting that.

pv:
PRs should be against daily, not master btw

dhermes:
@pv Thanks for the heads up.

PR #21, opened 2017-10-23
[config] use homebrew to install gfortran

xoviat:
There's no sense in maintaining more code than we need.


matthew-brett:
My worry as I remember it was that both the numpy and scipy wheels should be shipping the same gfortran libraries.  If we update with homebrew, the libraries will depend on what version homebrew had at the time.

xoviat:
Okay.

PR #24, opened 2018-01-27
Moved AppVeyor config file

radarhere:
Prepends dot to the AppVeyor config filename, as is supported - https://www.appveyor.com/docs/build-configuration/#yaml-file-alternative-naming

This makes the appearance of the config file more consistent with the .travis.yml file.


pv:
Thanks. At the moment I think this is not really necessary to do since this repo is just for building wheels.

radarhere:
Okay, thanks

PR #26, opened 2018-02-11
Update multibuild to the version that supports stripping wheels

### LINKED ISSUE #25: Update multibuild to a version that strips binaries in wheels
On the next release, please update the multibuild submodule to matthew-brett/multibuild@b2748e5 or newer. This should reduce the size of the scipy wheels from 44M to 30M by excluding unnecessary debug symbols. Also this updates to a newer version of `auditwheel` and `patchelf` that produce wheels with binaries that are actually strippable, i.e. fixes things like https://github.com/pypa/manylinux/issues/119

To make this work for `scipy` you need to `export FFLAGS="$FFLAGS -fPIC -Wl,-strip-all"` as the first line in `function build_wheel` in `config.sh`. For a successful Travis build see https://travis-ci.org/xhochy/scipy-wheels/builds/337663919

xhochy:
Fixes #25

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -10,6 +10,7 @@ OPENBLAS_VERSION=0.2.18
 source gfortran-install/gfortran_utils.sh
 
 function build_wheel {
+  export FFLAGS="$FFLAGS -fPIC -Wl,-strip-all"
     if [ -z "$IS_OSX" ]; then
         build_libs $PLAT
         build_pip_wheel $@


config.sh
```diff
@@ -10,6 +10,7 @@ OPENBLAS_VERSION=0.2.18
 source gfortran-install/gfortran_utils.sh
 
 function build_wheel {
+  export FFLAGS="$FFLAGS -fPIC -Wl,-strip-all"
```
pv: Is `-fPIC` right on 32-bit platforms?

pv: Also, adding `-Wl,-strip-all` seems to be included in multibuild https://github.com/matthew-brett/multibuild/commit/b2748e5d4d2ae6e45b130bdf8e24361083341e9a --- maybe we can bump the version to that commit instead and not set FFLAG here? Or maybe I'm missing something?

xhochy: `-fPIC` should work on all platform as far as I know.

That commit in `multibuild` only sets `-Wl,-strip-all` if no other `FFLAGS` are set, i.e. it supports that you can override the compile options if you don't want your things stripped. Thus we need to supply all flags here.

pv:
Thanks, seems to work fine.

Issue #30, opened 2018-05-04
Fortran shared libraries not stripped


pv:
gh-26 added stripping flags to FFLAGS, but it appears the so files linked by gfortran in manylinux wheels aren't actually stripped.

PR #31, opened 2018-05-05
Set LDFLAGS for stripping at link time

pv:
Numpy distutils appears to only look at the LDFLAGS environment variable
when determining how to link fortran.

Note that distutils and numpy.distutils interpret the LDFLAGS variable
in slightly different ways, with the latter requiring that -shared is
included there. <s>To work around this, we concatenate the Python and numpy
defaults.</s>

[Appveyor build cancelled manually, since this does not touch appveyor config.]

Fixes: #30


matthew-brett:
Thanks for this.

PR #32, opened 2018-06-04
MRG: build against OpenBLAS v0.3.0

matthew-brett:


matthew-brett:
Ok - thanks to the suggestion by @pv to use a recent numpy distutils, the macOS builds work correctly now, and the remaining travis failures are all spurious time-outs and so-on.  This one is ready to merge.

rgommers:
I'm not sure that patching in this repo is the right thing to do. If we need these changes to build on macOS, then we should apply them to `scipy/_build_utils`

rgommers:
Okay will do. Let's get this in to have things building again. Thanks @matthew-brett

PR #33, opened 2018-07-30
Build daily wheel on Python 3.7

naoyak:
Hi! The cron build for scikit-learn master which builds on SciPy daily master is not working correctly (https://github.com/scikit-learn/scikit-learn/issues/11697) since there isn't a daily wheel available for Python 3.7. Let me know if this looks right.


charris:
Multibuild is a submodule defined in a different repository
```
charris@fc [numpy-wheels.git (v1.15.x)]$ git submodule
 e9764d0a8f4e3adcd385c319e7742177193484c7 gfortran-install (remotes/origin/HEAD)
 afbff539ea815f3dafcd68f1bf1e87c4381bcbda multibuild (remotes/origin/HEAD)
 9e419a0359cb5dbb9ae0b2711ab7a59b121a98cf numpy (pre-removal-numpybook-6032-g9e419a035)
```
This is not the place to edit it. Note the the branch `v1.15.x` builds 3.7 wheels, you could just check for the difference. Note that nightly builds are not being run at this time, we need to enable cron builds and apparently the account that did that is gone. @matthew-brett What do you think is the best way forward?

naoyak:
> Note the the branch v1.15.x builds 3.7 wheels,

@charris are you referring to the numpy wheels rather than scipy? I thought I did the submodule update right...

charris:
Oops, yes, I was looking at numpy-wheels, not sure why I got notified for this. Anyway, you may have done the module update correctly, I can't tell by looking, but I don't recall seeing different files within the submodue listed as having changed, usually just the submodule itself is the only thing listed as changed. I assume that you changed to the submodule directory, did a git fetch, checked out the hash you wanted, then changed back to the head directory and made the commit.

matthew-brett:
Chuck - it's confusing, because the Github diff lists the files changed inside the submodule, when you update the submodule, but @naoyak did it correctly.

For some reason scipy-wheels was running monthly on the cron job, I've changed it to daily.

I'm not sure what's happening with Appveyor - I'm looking here:

https://ci.appveyor.com/project/scipy/scipy-wheels

Chuck - do you know who owns the Scipy account?  @pv - do you know?   Can you contact Appveyor to ask for Cron permission on that account?

pv:
Let's keep the crons weekly, I don't see particular reason for daily.

matthew-brett:
I prefer daily, partly because this kind of thing crops up from time to time:

https://github.com/scipy/scipy/issues/9091

pv:
Right, however, also taking that into account, weekly IMO is sufficient in view of the bi-yearly release cycle.

The issue with daily builds is that the MacPython and scipy queues tend to be full enough, so often when  do something, you have to wait.

pv:
It seems the testing part of the windows builds on appveyor doesn't work properly, #35

PR #37, opened 2018-11-14
MAINT: bump Cython to 0.29.0

tylerjereddy:
As [discussed](https://github.com/scipy/scipy/issues/9464#issuecomment-437541036) with @rgommers and @charris, bumping Cython version in preparation for  SciPy `1.2.0rc1` wheel building.

I've not messed with the Cython `language_level` directive stuff that now emits a FutureWarning -- leaving that alone seemed ok for generating the source distributions, though I don't know if that's bad practice.

Not sure if I'm going to need elevated permissions for this repo or not -- I think I'll need to create a branch for 1.2.x for wheel building soon, right? And point the wheel builds to the appropriate 1.2.x scipy repo branch rc commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
         - BUILD_COMMIT=master
         - PLAT=x86_64
         - NP_BUILD_DEP="numpy==1.8.2"
-        - CYTHON_BUILD_DEP="Cython==0.26.1"
+        - CYTHON_BUILD_DEP="Cython==0.29.0"
         - NP_TEST_DEP="numpy==1.8.2"
         - UNICODE_WIDTH=32
         - MANYLINUX_URL="https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,7 +16,7 @@ environment:
       OPENBLAS_64: "https://3f23b170c54c2533c070-1c8a9b3114517dc5fe17b7c3f8c63a43.ssl.cf2.rackcdn.com/openblas-v0.3.0-win_amd64-gcc_7_1_0.zip"
       OPENBLAS_32_SHA256: 4d8ebb7deb23e9aa39837087a8e41ac4cc0c0e44fa60861e1568e974bdbd7be6
       OPENBLAS_64_SHA256: e1bea2d94fe1c54c3cd1da62440a7975826427bc2dd55e67d064059632ddf323
-      CYTHON_BUILD_DEP: Cython==0.26.1
+      CYTHON_BUILD_DEP: Cython==0.29.0
       NUMPY_TEST_DEP: numpy==1.13.1
       TEST_MODE: fast
       APPVEYOR_SAVE_CACHE_ON_ERROR: true


mattip:
The `FutureWarning` will go away by itself in the next release cython. It will default to python3 constructs, the scipy python2 wheels may need to specify `cython -2` to work properly. 

Moving to cython 0.29 has the advantage that it should no longer emit the `size changed` warnings when building scipy with an older numpy version but running with a newer numpy version.

charris:
Yep, you will need commit privileges on MacPython/scipy-wheels. I don't know if anyone besides @matthew-brett can do that, maybe @rgommers or @pv. Note that you also need the updated Cython on you own machine in order to produce the source distributions.

You may also need to update the `gfortran-install` and `multibuild submodules`, the latter needed updating for the last `wheels` release. In any case, both differ from what I used for the last NumPy release.

charris:
The `travis` problem is due to the new `wheels` release, you need the submodule updates.

rgommers:
> Yep, you will need commit privileges on MacPython/scipy-wheels. I don't know if anyone besides @matthew-brett can do that, maybe @rgommers or @pv. 

Looks like I cannot add people to a team, but I can invite you as a collaborator with write permissions - just did so.

tylerjereddy:
Appveyor already had `git submodule update --init` -- I've updated to copy that command over to Travis config. Fingers crossed!

charris:
Hmm, the usual way to change the submodule version is somewhat complicated, IIRC

- cd to <submodule directory>
- git fetch origin
- git checkout <wanted hash or branch>
- cd root directory
- git add <submodule directory>
- git commit ...

There are variations on that, but the submodule directory acts like another repository. This way offers more control over what is in the submodule, don't know how your version works. I thought appveyor worked because it never referenced the parts of `wheel` that changed.

charris:
You probably want `upstream` instead of `origin` in the above. The numpy hashes are
```
 e9764d0a8f4e3adcd385c319e7742177193484c7 gfortran-install (remotes/origin/HEAD)
 951b6c64f01853cf2569000bb30ecd01a16bba0b multibuild (heads/devel)
 9e419a0359cb5dbb9ae0b2711ab7a59b121a98cf numpy (pre-removal-numpybook-6032-g9e419a035)
```
The current scipy are
```
 e9764d0a8f4e3adcd385c319e7742177193484c7 gfortran-install (remotes/origin/HEAD)
 436e764fe55fd5f07ca41404319126c759093694 multibuild (remotes/origin/add-install-new-python-6-g436e764)
 447e7e4b1c7c8257fba45da00f62f4bc14c693d3 numpy-distutils (pre-removal-numpybook-4820-g447e7e4b1)
 30749b5114e0dbc3439a12dba663865ba0f98fc8 scipy (v0.18.0rc1-13-g30749b511)
```
Don't forget to `git submodule update` before pushing your PR.

charris:
Not sure what `remotes/origin/add-install-new-python-6-g436e764` means for scipy. I think we should all be using the same multibuild. @matthew-brett ?

charris:
The pointers are stored in the root `git` and determine what you checkout when you run `git submodule update`, I think they should definitely be updated, I don't know of anyone doing it differently.

charris:
Also, you don't always want current upstream master, it may be broken. The point of having the version branches is to build all the micro releases the same way as much as possible, so that is another reason to update the hashes, they are branch specific.

charris:
Also, multibuild has two branches to worry about, `dev` and `master`, and the usable branch may differ.

charris:
If you look at the commit history you will see several updates to the multibuild submodule.

tylerjereddy:
Ok, I better try committing the sync for multibuild in this branch then. Perhaps I'll see if the hash recently used for NumPy works as a first pass.

charris:
Other failing tests on i386
```
FAIL sparse/tests/test_construct.py::TestConstructUtils::test_random_sampling
FAIL special/tests/test_kolmogorov.py::TestSmirnovp::test_oneovernclose
FAIL spatial/tests/test_distance.py::TestPdist::test_pdist_jensenshannon_iris_nonC
```

tylerjereddy:
Hmm, the Travis wheels cron job has been failing for the past 10 weeks so naturally things have popped up I suppose.

Some of those issues / tracebacks are clearly related to recently-merged SciPy PRs. I guess the toolchain/ ecosystem around the wheel building process is such that it samples scenarios not tested in conventional SciPy development CI, unfortunately.

charris:
Yeah, I was surprised to see all the failing cron jobs, due to the wheels problem I think. I get notices from travis when the NumPy cron job fails.

charris:
The other failing tests may simply be precision problems. Note that OpenBLAS 0.2.15-0.3.3 all have threading problems and I think you are using 0.2.18. I'm thinking of delaying NumPy 1.16 until 0.3.4 comes out if it isn't too far off.

charris:
Some of the other failures look Python version dependent, 3.6 and 3.7 64 bits look OK.

tylerjereddy:
I was able to "fix" the `test_random_sampling` failure locally as described in the cross-referenced SciPy PR, but will have to wait for some feedback on the approach we actually want to take & backporting suitability.

rgommers:
Okay, so two other ones besides the `random_sampling` one (not in all test runs):
```
FAIL special/tests/test_kolmogorov.py::TestSmirnovp::test_oneovernclose
FAIL spatial/tests/test_distance.py::TestPdist::test_pdist_jensenshannon_iris_nonC
```
See e.g. https://api.travis-ci.org/v3/job/455687609/log.txt. The `test_kolgomorov` one looks annoying, the other one seems to only need a tiny bump in test tolerance.

tylerjereddy:
I'll hold off on additional activity here until the remaining i386 issue fix / test skip is merged & the 64-bit sparse issue is fixed (somehow, or PR reverted) in SciPy master.

Then I'll proceed with checking to see if this PR is all-green using SciPy master & use that as (partial?) basis for presumably justifying backports to the 1.2.x maintenance branch for the 3 failures.

Technically, it is possible to follow the release instructions that allow (indeed, suggest) pushing the REL commit **before** checking on the wheels (only source builds happen before that commit), so I'm guessing some clean-up will be needed once this is sorted out--otherwise I'll have 1.2.0rc1 REL commit preceding useful backports and not matching the commit hash I'll want to point the wheel builds toward.

charris:
I went ahead a merged as we might as well have the wheel and cython updates in the master branch for the weekly/test builds.

charris:
Thanks Tyler. Since the test fixes go into scipy master, I don't see a problem with putting this in now.

charris:
BTW, for testing you can just push directly upstream with updated scipy commit hash. When you get ready to release, make a `v1.2.x` branch, set the relevant commit hashes and push the new branch.

PR #42, opened 2018-12-24
Update SciPy v1.2.x to OpenBLAS 3.4+

charris:
Update to the fixed up v0.3.4 version, which is sadly named as being
v0.3.3. The newer library fixes many threading problems found in the
earlier version as well as other bugs, in particular https://github.com/scipy/scipy/issues/9620.

Note that this PR also upgrades the OS X deployment target from 10.6 to
10.9, which is needed because the new OpenBLAS libraries target that
version.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,10 +12,10 @@ environment:
   global:
       MINGW_32: C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\bin
       MINGW_64: C:\mingw-w64\x86_64-6.3.0-posix-seh-rt_v5-rev1\mingw64\bin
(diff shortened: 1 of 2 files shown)


tylerjereddy:
@rgommers Do you think the bug fix value exceeds the possible issues caused by switching openblas versions used to build wheels between `1.2.0` & `1.2.1`.

Perhaps worth it given that this is an LTS branch & more work is going to be hard to avoid anyway.

The openblas version confusion may eventually be relieved by post-mortem version inspection on the wheels using my related work in NumPy, but out of scope for this maintenance branch.

rgommers:
Yes, I think it's worth it, the reported bug alone is worth fixing, and there were all the threading issues as well.

And agreed not to worry about version number thing here.

tylerjereddy:
Ok, in it goes, thanks @charris

PR #43, opened 2019-01-04
Add std=c99 flag to wheel build.

charris:
This is needed for the local declaration of loop variables. SciPy
does not do that currently, but this guarantees that if
they are used in the future a compiler error will not be raised.

Might also want to add `-fno-strict-aliasing` to the flags.

### DIFF
--- a/env_vars.sh
+++ b/env_vars.sh
@@ -1,6 +1,7 @@
 # Environment variables for build
 OPENBLAS_VERSION="v0.3.3-186-g701ea883"
 MACOSX_DEPLOYMENT_TARGET=10.9
+CFLAGS="-std=c99"
 
 # Enable Python fault handler on Pythons >= 3.3.
 PYTHONFAULTHANDLER=1


env_vars.sh
```diff
@@ -1,6 +1,7 @@
 # Environment variables for build
 OPENBLAS_VERSION="v0.3.3-186-g701ea883"
 MACOSX_DEPLOYMENT_TARGET=10.9
+CBLAS="-std=c99"
```
rgommers: `CBLAS` should be `CFLAGS` I think?

charris:
Interesting. This is causing errors on the i386 platform. I'm also seeing errors for NumPy on i386.

rgommers:
Hmm, that doesn't look like fun.

pv:
Closing (old PR). Similar flags would need to be added to the MSVC build. I guess we're going stick with C89 like dialect for a while longer...

PR #44, opened 2019-04-09
MAINT: bump min Cython version

tylerjereddy:
Our master branch scipy wheels Travis cron jobs are failing because the pin to Cython `0.29.0` no longer meets our minimum requirement. For those jobs that are pinned to the minimum version, this PR should bump to the one we now require.

We don't have Appveyor cron jobs as far as I can tell, but there's a similar setting there that I've changed in the same way.


rgommers:
Merged, thanks Tyler

PR #46, opened 2019-07-07
MAINT: add pybind11 build dependency

pv:
Due to the addition of pocketfft to scipy, pybind11 is now a build dependency.


pv:
Ok, straightforward and seems to work, so merging.

PR #47, opened 2019-08-11
MAINT: remove some old Python content

tylerjereddy:
Noticed this based on code highlighted by Pauli here: https://github.com/scipy/scipy/issues/10628

A follow-up PR may be needed if a decision is reached on the Python 3.7+ issue described there.

* remove some comments and code related
to versions of Python we no longer support;
the LTS Python 2.7 branch has its own wheels
branch that won't be affected by this

I suspect even more cleanup is possible depending on the 3.5 support timeline, but perhaps this is a conservative start.


rgommers:
LGTM, merged. thanks Tyler

PR #51, opened 2019-10-12
Fix py35-osx build on travis

pv:
The build fails when trying to install pytest-forked included via
TEST_DEPENDS, when it tries to easy-install via setup_requires the
setuptools_scm package, encountering some TLS download errors.  Work
around this by pre-installing the package via pip.


rgommers:
Merged, thanks Pauli

PR #52, opened 2019-10-28
MAINT: bump Cython for Travis wheels

tylerjereddy:
* four of the master branch SciPy wheels
Travis CI matrix entries are erroring with
a message: "Building SciPy requires Cython >= 0.29.13, found 0.29.2"

* so, bumping Cython to 0.29.13 in an attempt to
fix this error

Note that I'd still expect those jobs to fail, but SciPy should at least build if this helps. There appears to be another issue even if Cython is updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
         - BUILD_COMMIT=master
         - PLAT=x86_64
         - NP_BUILD_DEP="numpy==1.13.3"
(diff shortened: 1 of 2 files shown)


rgommers:
okay worked after updating `appveyor.yml` with the same change. failures are all `stats.histogram_dd`, which is unrelated. merged, thanks Tyler

PR #55, opened 2019-10-29
WIP, MAINT: add 3.8 wheels

tylerjereddy:
* add Linux, MacOS, and Windows Python
3.8 wheel builds, largely inspired by
the changes used to generate 3.8 wheels
for NumPy

cc @charris @mattip


tylerjereddy:
The only job that passed was Linux Python 3.8 64-bit. The `TestBinnedStatistic.test_dd_zero_dedges` already has a fix in the works.

Here's a summary of the other test failures:

- most 32-bit jobs are showing a new failure `TestUpfirdn.test_modes`
- Mac Python 3.6 is showing a new failure `TestTf2zpk.test_simple`
- Mac Python 3.7 has a build failure that looks a bit like the one Pauli addressed in master by pinning multibuild version: `ar: build/temp.macosx-10.6-intel-3.7/libsuperlu_src.a: Inappropriate file type or format`
  - I think Matti suggested setting multibuild to the HEAD of its develop branch, but maybe we can't do that for SciPy & should aim to pin like Pauli did in main repo?
- Mac & Windows Python 3.8: `ImportError: cannot import name 'pickle' from 'numpy.compat' (/Users/travis/build/MacPython/scipy-wheels/venv/lib/python3.8/site-packages/numpy/compat/__init__.py)`

mattip:
Can you set the macOS builds to use MB_PYTHON_OSX_VER=10.9 ? If you do not use HEAD of multibuild devel, you cannot build on python3.8

mattip:
The pickle error seems to be some kind of circular import problem, that resolves differently on python3.8

grlee77:
The failure in `TestUpfirdn.test_modes` is related to a recent PR I made and appears to just be an issue of relaxing the tolerance slightly.  I think just dropping rtol/atol from 1e-6 to 1e-5 will probably fix it. I will make a PR.

(`TestUpfirdn.test_modes` should not be in the 1.3.x branch, so it shouldn't hold up a 1.3.2 release)

eric-wieser:
Is the bug that `compat.py` doesn't define pickle on python 3.8?

mattip:
[compat.py](https://github.com/numpy/numpy/blob/v1.17.3/numpy/compat/py3k.py#L20) should raise if it can't find pickle, and that is passed back out in the [`__init__.py`](https://github.com/numpy/numpy/blob/v1.17.3/numpy/compat/__init__.py#L16)

tylerjereddy:
> Is the bug that compat.py doesn't define pickle on python 3.8?

Hmm, I simply can't reproduce in a pretty clean virtualenv with Python 3.8 on a Mac--it seems other devs are having similar experiences in the cross-linked NumPy PR

eric-wieser:
> Here's a summary of the other test failures:

Can you add links to the failure logs for each?

eric-wieser:
Is [this](https://ci.appveyor.com/project/scipy/scipy-wheels/builds/28471652/job/r4hafeiiivyjy1cb#L252) the problem?
```
rm -r -Force "$NumpyDir\compat"
mv numpy-distutils\numpy\compat $NumpyDir
```

charris:
@eric-wieser I was looking at numpy-distutils also. It points to an old version of NumPy among other things.

appveyor.yml
```diff
@@ -149,11 +163,14 @@ install:
 
   # Replace numpy distutils with a version that can build with msvc + mingw-gfortran.
   - ps: |
+      $PYTHON_VERSION = $env:PYTHON_VERSION
+      If ($PYTHON_VERSION -ne 3.8) {
```
tylerjereddy: In the most recent commit I've added this guard to see if blocking the distutils override/patching can allow Appveyor to at least build SciPy with Python 3.8 on Windows (32 & 64-bit).

This has been in various parts of our CI for quite some time because of a long-standing distutils limitation for the lower versions we supported I think.

It would probably be more precise to put the guard on the version of NumPy, but this should suffice for testing at least.

tylerjereddy:
That was helpful--SciPy appears to be building for Python 3.8 on Windows/Appveyor now. Next problem is that the matplotlib version used on Appveyor is too old for 3.8, so I've bumped that to the release candidate version we now use in the main repo CI.

charris:
I think you can simply update the `numpy-distutils` submodule. The original PR #4 that added it did so to obtain fixes from numpy master that apparently were not in a released version.

tylerjereddy:
Now a bunch of errors in SciPy Python 3.8 testsuite on Appveyor of this nature:

`ImportError: DLL load failed while importing ...`

tylerjereddy:
IIRC there were some DLL location handling adjustments in recent distutils, with the specification of that location eventually abstracted out to a distributor init file, though I'm not sure that is the problem here.

tylerjereddy:
I added the `numpy-distutils` submodule update to latest NumPy hash available at time of writing. That probably means I can back out the Appveyor distutils swap guard--for now, let's just see if that fixes the Mac Python 3.8 SciPy build issue (it should), since the distutils swap there was separate & I didn't touch it yet.

tylerjereddy:
I think @mattip may try to reproduce the Windows DLL issue locally. I won't have access to a Windows machine until I return from the summit in NY.

We don't have to get this PR all green right away--will want it reasonably clean so that a similar PR can be opened to the 1.3.x wheels branch though, so that 1.3.2 Py 3.8 support can be available soon (and some CI fails are from master changes that aren't on that branch).

tylerjereddy:
Mac Python 3.8 is now passing--nice.

tylerjereddy:
Trying a distribution file to assist with SciPy Windows DLL loading. A bit speculative that this is the actual problem, but based on the approach from https://github.com/numpy/numpy/pull/13019

tylerjereddy:
I've adjusted the appveyor code to copy the distribution file *after* the `git checkout` of scipy---hopefully that will alleviate the issue currently observed. I'm a little slow at the moment because of the travel/recovery from NumFOCUS meeting.

tylerjereddy:
DLL issue persists for Windows Python 3.8--I'll probably have to debug locally if I can manage to reproduce, but not surprising given the DLL resolution changes in that Python version.

The remaining Windows issues are less serious it seems -- just need to guard against the matplotlib version upgrade for older Python 3.5 & a master branch fix is incoming for the stats test (doesn't affect 1.3.2 anyway).

tylerjereddy:
oh, I see `mv : Cannot create a file when that file already exists` -- yay windows! So the DLL fix may yet work without too much hassle..

tylerjereddy:
Ok, I've tried to patch that DLL issue by removing the target distributor init file first, and then moving the new one in.

I've also added a guard for the new matplotlib rc version--only really needed for 3.8 & 3.5 doesn't work with it.

tylerjereddy:
The problematic Windows wheel can be downloaded directly from the Appveyor artifacts page, such [as here](https://ci.appveyor.com/project/scipy/scipy-wheels/builds/28707252/job/ltyydmuwsefnn7h6/artifacts). That should simplify debugging--I'll take a quick look to see if I can at least reproduce.

tylerjereddy:
Reproduced the 173 DLL errors locally on Windows/powershell/python 3.8 using that wheel artifact download and these steps/notes:

- working at path: `C:\Users\treddy\rough_work\scipy\wheel_debug`
- `pip install scipy-1.4.0.dev0+f1648dd-cp38-cp38-win_amd64.whl`
- the wheel was installed to path: `C:\Users\treddy\windows_dev_work\conda_install\envs\scipy_dev_py38\lib\site-packages\scipy`
- `cd C:\Users\treddy\github_projects\scipy-wheels\scipy`
- `python ..\run_scipy_tests.py full`
- `173 error in 10.92s` (they look the same.. DLL issues)

That should provide some traction---@rgommers if you want to debug feel free since I've been slow. Otherwise, I'll likely continue this evening/over long weekend.

I suspect we can try to fix locally by iterating on a `_distributor_init.py` injected directly into the appropriate  path within the `whl` file, probably starting by using the version I wrote in this PR & adding debug code to it.

charris:
@tylerjereddy I've been holding off on releasing NumPy 1.17.4 in case it needs fixes for SciPy on 3.8. Do you think NumPy changes are likely needed or should I just go ahead with the release?

rgommers:
@charris if you can hold off for a few days, that may be useful. we'll try to get this sorted out on the weekend.

_distributor_init.py
```diff
@@ -0,0 +1,26 @@
...
+# on Windows SciPy loads important DLLs
+# and the code below aims to alleviate issues with DLL
+# path resolution portability with an absolute path DLL load
+if os.name == 'nt':
+    from ctypes import WinDLL
+    import glob
+    # convention for storing / loading the DLL from
+    # scipy/.libs/, if present
+    libs_path = os.path.abspath(os.path.join(os.path.dirname(__file__),
+                                             '..', '.libs'))
```
mattip: since this file is in the top level directory, you do not need the `'..'`. Removing that from my local copy let
`python -c"import scipy.linalg"` succeed where with this it fails

mattip: Tests pass too
<details>

```
================================================ warnings summary =================================================
Python38/Lib/site-packages/scipy/stats/tests/test_binned_statistic.py::TestBinnedStatistic::test_dd_zero_dedges
  c:\pypy_stuff\Python38\lib\site-packages\scipy\stats\_binned_statistic.py:699: RuntimeWarning: divide by zero encountered in log10
    decimal = int(-np.log10(dedges[i].min())) + 6

-- Docs: https://docs.pytest.org/en/latest/warnings.html
====== 19288 passed, 1305 skipped, 10954 deselected, 78 xfailed, 8 xpassed, 1 warnings in 391.50s (0:06:31) =======
```

tylerjereddy: Thanks Matti--I've pushed a commit with you as co-author. We should be much closer overall now to things working.

charris:
You can speed up the testing by commenting out most of the builds.

tylerjereddy:
Remaining failures look like some subtle test things in a few cases for Mac (Python 3.5/3.6) & the more annoying SUPERLU thing for mac Python 3.7.

Useful to keep in mind that some of the remaining issues may be master branch-specific so likely worth testing in the 1.3.2-specific #56, since we can have a little more time before the master/1.4.x wheels branch.

tylerjereddy:
The only failure that overlaps with the 1.3.x wheels branch in #56 is the superlu-related Python 3.7 build failure on mac.

Once that is fixed, I suspect we are good to go & I will just cherry-pick the appropriate commits from here.

tylerjereddy:
Since the latest master branch of multibuild is only needed for Python 3.8, I could try to cheat and pin multibuild as @pv did for our main repo CI here: https://github.com/scipy/scipy/pull/10917/files

I'm not sure how unusual it would be to use different multibuild versions for different Travis entries, and probably not ideal from a maintenance/predictability standpoint, but I may try it to move things forward.

tylerjereddy:
I've added an attempt to pin multibuild for MacOS/Python 3.7. I've also reduced the size of the test matrices substantially while we try to sort that out as Chuck suggested.

tylerjereddy:
Ok, the superLU issue has disappeared--I'll make an attempt to cherry-pick relevant commits to #56 now.

tylerjereddy:
I've merged #56 for the 1.3.2 wheels branch as it was all green for similar changes. I'll return to this PR once I get 1.3.2 out the door. For this PR, we could probably just revert the debug commit that reduces the CI matrix & the improvements would be ready to merge with Mac Travis failures fixed in a follow-up PR or in master branch/main repo.

tylerjereddy:
I've reverted the debug matrix reduction commit--if CI looks solid after that (just 3 Travis Mac issues), this is probably ready to go as we look toward 1.4.0, modulo any reviewer comments here?

rgommers:
The only test failure in three of the build matrix entries is `TestTf2zpk.test_simple[complex128]`, which is unrelated to this PR. The 3.8 entries look good, so merging. Thanks @tylerjereddy

rgommers:
Following up on test failure in https://github.com/scipy/scipy/issues/11041

PR #60, opened 2019-12-17
BLD: update to pybind11 2.4.3, fixes a segfault with SciPy 1.4.0

rgommers:
See https://github.com/scipy/scipy/issues/11237 and
https://github.com/pybind/pybind11/issues/1262 for details.

tl;dr pybind11 is the first version where its symbol names are prefixed
with the GCC (or Clang) version it was compiled with.

PR #61, opened 2019-12-17
BLD: update to pybind11 2.4.3, fixes a segfault with SciPy 1.4.0

tylerjereddy:
Wheels repo backport of #60 in preparation for SciPy `1.4.1`.


rgommers:
Merged, thanks @tylerjereddy. No need for this simply commit to run 5x for every PR, every merge, and then again for v1.4.1

pv:
The openblas-libs rebuild just a moment ago changed openblas hashes, so they'd need to be updated.

rgommers:
Is that for the 64-bit BLAS? If so, wouldn't it be better to leave that out of v1.4.1?

tylerjereddy:
The regular OpenBLAS got rebuilt as well as a side-effect I think, because I could not cancel those jobs. I probably should have just deleted the matrix entries but I was busy I guess & want to quickly provide those for the NumPy testing stuff.

pv:
Yes, the openblas-libs setup is a bit dangerous vs. overwriting existing artifacts (and several projects rely on those). Could maybe be better to add some "dev" env flag on the master branch, and forbid the wheel uploader from replacing artifacts on non-dev builds.

tylerjereddy:
In addition to the hashes I also see one fail for `scipy.optimize.tests.test__differential_evolution.TestDifferentialEvolutionSolver`.

PR #62, opened 2019-12-18
MAINT: update OpenBLAS sha256

tylerjereddy:
* upstream work in openblas-libs accidentally
regenerated at least a few of the zip files
we use from rackspace for OpenBLAS v0.3.7 in wheel
build workflows; adjust the sha256 in those
cases where failures were recently observed
due to mismatches

Reference log with only 32-bit Windows sha256 mismatches: https://ci.appveyor.com/project/scipy/scipy-wheels/builds/29611388


tylerjereddy:
The jobs that were previously failing because of the sha256 hash change of 32-bit windows OpenBLAS are now passing again. Travis CI failures are not related and an issue is already open on the main repo.

Remaining entries are green, with one queued appveyor entry (appveyor is getting hammered here with all the dev work going on in MacPython I think).

So, merging.

PR #65, opened 2020-01-23
MAINT: add support for pip 20

tylerjereddy:
* sync multibuild submodule to latest `devel`
branch to pull in fixes for supporting pip `20`;
this is currently causing several new failures
in Travis CI of the wheels repo

Upstream reference is: https://github.com/matthew-brett/multibuild/pull/299

I believe @charris has already done this for `numpy-wheels`.

Example Travis CI fails: https://travis-ci.org/MacPython/scipy-wheels/builds/640278508
Note that if this patch works, some jobs will still fail for the same reasons as before, but several matrix entries should pass again.

Appveyor currently appears unaffected: https://ci.appveyor.com/project/scipy/scipy-wheels/builds/30275123


tylerjereddy:
This almost restores the Travis CI matrix to the state of the last cron job, which is an improvement, so I'll merge.

There is one exception, which is that one of the Mac OS jobs, with Python 3.7, is still pinned to an older multibuild for the same reason this is done in the master repo for SciPy development proper.  As noted in #55 

> Mac Python 3.7 has a build failure that looks a bit like the one Pauli addressed in master by pinning multibuild version: ar: build/temp.macosx-10.6-intel-3.7/libsuperlu_src.a: Inappropriate file type or format
I think Matti suggested setting multibuild to the HEAD of its develop branch, but maybe we can't do that for SciPy & should aim to pin like Pauli did in main repo?

I'll check that in a follow-up --- I think there are a few things we may need to clean up on the MacOS side of things, possibly including some naming issues as well.

PR #68, opened 2020-02-16
WIP: Anaconda org weekly upload

tylerjereddy:
- Don't merge yet, debugging...

* draft code to switch weekly Travis CI
SciPy wheels uploads to the new anaconda
cloud infrastructure; temporarily disable
the cron requirement so we can see if it
actually works

* secret token was added separately in Travis
CI settings

* debug commit to reduce CI matrix size while debugging


tylerjereddy:
I suspect I'll need to install `conda` in Travis first.

There's probably a linting failure if Travis isn't even showing in the matrix.

tylerjereddy:
Will have to reopen as a PR from the repo itself to access secret variable I think.

PR #71, opened 2020-05-09
MAINT: appveyor rack->conda.org

tylerjereddy:
* switch to using `anaconda.org` for OpenBLAS downloads
in our Windows wheel build infrastructure to avoid future
charges from Rackspace to OSS community

* this involves offloading most of the download logic
to `tools/openblas_support.py`

@mattip @pv @rgommers Any security concerns with removal of the hash checking here? Does that need to be added to `openblas_support.py` in the main repo?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,10 +12,6 @@ environment:
   global:
       MINGW_32: C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\bin
       MINGW_64: C:\mingw-w64\x86_64-6.3.0-posix-seh-rt_v5-rev1\mingw64\bin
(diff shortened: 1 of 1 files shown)


rgommers:
Is it just security, or is it also that we want to know exactly when we bundle an updated version rather than it just happening? I'd imagine at least in a bug fix release we don't want it to change unless there's a known issue.

mattip:
+1 for hash checking via hardcoded hashes in openblas_support.py

tylerjereddy:
ok thanks

tylerjereddy:
CI results look good now, I think

tylerjereddy:
The remaining CI fails are unrelated `signal` issues Ralf will fix later, so merging so that I can continue the migration effort.

PR #74, opened 2020-05-13
MAINT: appveyor conda.org uploads

tylerjereddy:
* turn on conda.org uploads for appveyor CI;
we never did weekly uploads for Appveyor/Windows
wheels so just turn on for merge events

* hopefully this will restrict to merges and not
successful PR branch runs by virtue of secret variable
availability only within the repo itself (so most PRs
should be opened from forks where possible)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -212,12 +212,11 @@ after_test:
   - C:\cygwin\bin\du -hs "%LOCALAPPDATA%\pip\Cache"
 
 on_success:
-  # Upload the generated wheel package to Rackspace
+  # Upload the generated wheel package to anaconda.org
+  # SCIPY_STAGING_UPLOAD_TOKEN is an encrypted variable
+  # used in Appveyor CI config, originally created at
+  # multibuild-wheels-staging site
   - cd ..\scipy
-  - pip install wheelhouse-uploader
-  # disable uploads because we have lost free hosting
-  # with Rackspace:
-  #- python -m wheelhouse_uploader upload
-  # --local-folder=dist
-  # %UPLOAD_ARGS%
-  # %CONTAINER%
+  - cmd: set ANACONDA_ORG="multibuild-wheels-staging"
+  - pip install git+https://github.com/Anaconda-Server/anaconda-client
+  - IF NOT "%SCIPY_STAGING_UPLOAD_TOKEN%" == "" anaconda -t %SCIPY_STAGING_UPLOAD_TOKEN% upload --force -u %ANACONDA_ORG% "%APPVEYOR_BUILD_FOLDER%/wheelhouse/*.whl"


appveyor.yml
```diff
@@ -212,12 +212,11 @@ after_test:
...
-  - pip install wheelhouse-uploader
-  # disable uploads because we have lost free hosting
-  # with Rackspace:
-  #- python -m wheelhouse_uploader upload
-  # --local-folder=dist
-  # %UPLOAD_ARGS%
-  # %CONTAINER%
+  - cmd: set ANACONDA_ORG="multibuild-wheels-staging"
+  - pip install git+https://github.com/Anaconda-Server/anaconda-client
+  - anaconda -t %SCIPY_STAGING_UPLOAD_TOKEN% upload --force -u %ANACONDA_ORG% "%APPVEYOR_BUILD_FOLDER%/wheelhouse/*.whl"
```
tylerjereddy, line 222: appveyor is complaining about this last line.. maybe something related to the variable substitution syntax is wrong?

tylerjereddy: actually, I probably just need to guard for the non-existence of `SCIPY_STAGING_UPLOAD_TOKEN` in feature branches before the merge happens.

tylerjereddy:
CI finally looks good now--let's see if the appveyor staging uploads happen after a merge event.

mattip:
The weekly wheels have a git hash in their name, which does not indicate chronological order. The old multiwheel uploaded added the date to allow ordering the uploads.

In numpy-wheels I added a rename function to add the date in the [weekly uploads](https://anaconda.org/scipy-wheels-nightly/scipy/files). That way pip will take the latest. It changes ` scipy-1.5.0.dev0+6f0b508-cp38-cp38-macosx_10_9_x86_64.whl` into ` scipy-1.5.0.dev0+20200515101522_6f0b508-cp38-cp38-macosx_10_9_x86_64.whl. 

The staging wheels do not need that, since they do not have a git-commit hash in their name.

tylerjereddy:
> The weekly wheels have a git hash in their name, which does not indicate chronological order. The old multiwheel uploaded added the date to allow ordering the uploads.

The conda.org site has a sorting column you can click based on upload timestamp. Not saying I'm against adding a datestamp to the weekly Travis uploads for SciPy though. May depend on how other projects consume them I guess.

Will `pip` check the timestamp metadata on binaries if we don't change the filenames?

> The staging wheels do not need that, since they do not have a git-commit hash in their name.

Ours do have a git commit hash in their name, until the release proper. So on a merge to master the staging upload area is populated with the dev version and hash. On a release feature branch the hash would presumably be excluded to give a pure "release version."

Something is going to go wrong for the next release attempt, I can feel it... :)

tylerjereddy:
So, for staging uploads:

1) for final releases, yes should only have version number and are sorted that way by pip I guess
2) otherwise, the staging area basically just contains "junk" from merges for debugging/the release manager, so the hashes without dates are probably fine there

mattip:
MacPython/numpy-wheels#84 was motivated by pandas-dev/pandas#33968. The rackcdn locations do have a date http://66ce22ef0bf26a5e2cca-4ffdece11fd3f72855e4665bc61c7445.r49.cf2.rackcdn.com/, and https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com/.

tylerjereddy:
I see, I guess that's an informal indication that `pip` may not be automagically checking timestamps.

PR #75, opened 2020-05-14
MAINT: Travis upload syntax

tylerjereddy:
* Travis wheel uploads, both for weekly and merge
events, are failing with an apparent syntax issue:

* push event log example fail:
https://travis-ci.org/github/MacPython/scipy-wheels/jobs/686382191

* cron event log example fail:
https://travis-ci.org/github/MacPython/scipy-wheels/jobs/686383899

* looks like a `;` is missing at the end of some `pip install` lines
in conditional blocks, so add those in to prevent the `pip install`
command from propagating to the next line


tylerjereddy:
It doesn't affect appveyor and Travis results look "ok", so let's try it.

PR #79, opened 2020-05-19
MAINT: use `openblas_support.py` for Travis

### LINKED ISSUE #72: OpenBLAS usage with Travis
I believe our wheel builds on Travis may be using some of the upstream submodule (multibuild) code to do some of the OpenBLAS handling.

I think we want to move to the `openblas_support.py` infrastructure for Travis as we just did for Appveyor.

This looks a bit less trivial for Travis though, since we depend on some upstream code here. Maybe not too bad though--what did you do for NumPy @mattip @charris ? I think you adopted a completely new approach with Azure, but I'm not sure I want to dive into that for the current release at this stage.

cc @rgommers 

I've added a comment in a multibuild repo issue since I also noticed some leveraging for `rackcdn` there still: https://github.com/matthew-brett/multibuild/issues/304#issuecomment-626397103

tylerjereddy:
Fixes #72 

* Appveyor CI wheel builds already use `openblas_support.py`; use the
same script for `OpenBLAS` dependency handling in Travis CI for
wheel builds

* this approach is based on the one described by @mattip for
NumPy wheels in the cognate issue, except that for SciPy we should
not need `urllib3` because of our different approaches to file
downloading in `openblas_support.py`

@rgommers @pv Note that this means `OpenBLAS` version bumps have to happen through merges to the maintenance branch via changes to `openblas_support.py` rather than setting an ENV variable in the wheels repo to bump the version.

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -17,12 +17,11 @@ function build_wheel {
 }
 
 function build_libs {
-    local plat=${1:-$PLAT}
-    local tar_path=$(abspath $(get_gf_lib "openblas-${OPENBLAS_VERSION}" "$plat"))
-    # Sudo needed for macOS
-    local use_sudo=""
-    [ -n "$IS_OSX" ] && use_sudo="sudo"
-    (cd / && $use_sudo tar zxf $tar_path)
+    PYTHON_EXE=`which python`
+    $PYTHON_EXE -c"import platform; print('platform.uname().machine', platform.uname().machine)"
+    basedir=$($PYTHON_EXE scipy/tools/openblas_support.py)
+    $use_sudo cp -r $basedir/lib/* /usr/local/lib
+    $use_sudo cp $basedir/include/* /usr/local/include
 }
 
 function set_arch {
--- a/env_vars.sh
+++ b/env_vars.sh
@@ -1,5 +1,4 @@
 # Environment variables for build
-OPENBLAS_VERSION="v0.3.7"
 MACOSX_DEPLOYMENT_TARGET=10.9
 
 # Enable Python fault handler on Pythons >= 3.3.


config.sh
```diff
@@ -17,12 +17,11 @@ function build_wheel {
...
-    local tar_path=$(abspath $(get_gf_lib "openblas-${OPENBLAS_VERSION}" "$plat"))
-    # Sudo needed for macOS
-    local use_sudo=""
-    [ -n "$IS_OSX" ] && use_sudo="sudo"
-    (cd / && $use_sudo tar zxf $tar_path)
+    PYTHON_EXE=`which python`
+    $PYTHON_EXE -c"import platform; print('platform.uname().machine', platform.uname().machine)"
+    basedir=$($PYTHON_EXE scipy/tools/openblas_support.py)
+    $use_sudo cp -r $basedir/lib/* /usr/local/lib
+    $use_sudo cp $basedir/include/* /usr/local/include
```
tylerjereddy, line 24: we'll see if I can get away with almost copy-pasting the solution from `numpy-wheels`...

tylerjereddy:
CI is all green and I spot checked one of the Travis logs, which shows:

`Downloading: https://anaconda.org/multibuild-wheels-staging/openblas-libs/v0.3.9/download/openblas-v0.3.9-manylinux1_x86_64.tar.gz`

Which is exactly what we want.

So, I'll go ahead and merge this and open separate issues to remind me to:
- spot check the OpenBLAS versions for the uploaded wheels after this gets merged
- possibly use the `openblas_support.py` version checking machinery directly for the POSIX/Travis CI wheel builds, though this is easiest when the package is actually installed, I think

PR #83, opened 2020-05-30
MAINT: bump Cython to min required

tylerjereddy:
* the minimum required Cython version to build
SciPy is now `0.29.18`; this change has been pushed
directly to the `v1.5.x` wheels feature branch because
of observed failures there, so this is effectively
a forward-port of that


tylerjereddy:
Simple maintenance to keep wheels repo builds going and CI all green, self-merging.

Issue #77, opened 2020-05-15
MAINT: Timestamp nightly wheel filenames


tylerjereddy:
Based on discussion with @mattip in #74, downstream projects using `pip` to pull in our wheels from the nightly (weekly) anaconda.org repo appear to need dates stamped right into filenames, presumably because `pip` is not scraping timestamp metadata  directly from the uploaded files.

So it just kind of arbitrarily sorts by hash alphanumerically, which is bad.

While I'd prefer if there were a simple `pip install` flag of some sort that "just worked" for pulling in the latest based on metadata, I'll open this issue to try to get done soon-ish.

At least for us, we only need to worry about Travis--we haven't done nightlies for Windows/Appveyor in recent history AFAIK.

bashtage:
I don't think you have to timestamp.  You could also use the structure

```
scipy-1.5.0.dev0+NUMBER_OF_COMMITS_AFTER_dev0.g6fd3052-cp38-cp38-macosx_10_9_x86_64.whl
```

tylerjereddy:
Following the merging of the cross-linked PR above--I should check mid-weed after the Tuesday "nightly" upload to see if the wheels are properly named for downstream `pip` download purposes.

tylerjereddy:
The timestamping appears to have worked as intended for the weekly uploads last Tuesday: https://anaconda.org/scipy-wheels-nightly/scipy/files

I've also manually purged the older weekly wheel uploads, with incorrect file naming, to prevent interference with automated downloads from downstream projects using `pip` in CI, etc.

So, I'll close this issue. @bashtage -- feel free to open an issue/ping me if `statsmodels` or other projects are having issues with our weekly wheels.

PR #84, opened 2020-05-30
MAINT: timestamp nightlies

tylerjereddy:
* as outlined in gh-77, downstream projects will have
a hard time automatically downloading our most recent
"nightly" (actually weekly) wheel uploads if we don't
add a timestamp of sorts

* this patch attempts to "borrow" the machinery added by
Matti in https://github.com/MacPython/numpy-wheels/pull/84
for this purpose

I believe this has already impacted one project for @bashtage

cc @mattip @rgommers 

I don't want this to automatically `close` gh-77 because we should check the next nightly upload filenames once this is ready & merged (weekly uploads happen on Tuesdays, for POSIX platforms only)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,6 +120,8 @@ after_success:
     # used in Travis CI config, originally
     # generated at anaconda.org for scipy-wheels-nightly
     - if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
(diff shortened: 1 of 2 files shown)


tylerjereddy:
CI all green + there's an approval--let's try it & I'll make a note in the matching issue to check mid-week to see if it worked.

PR #88, opened 2020-06-28
MAINT: DLL load adjustments

tylerjereddy:
* there were numerous reports of downstream
DLL loading issues in the SciPy main repo
issue https://github.com/scipy/scipy/issues/12400

* follow some of the suggestions there, including
usage of `__config__` earlier in the SciPy loading
control flow and bumping the version of the
`numpy-distutils` submodule

* consistently reproducing the original issue has
proven tricky, even working locally on a native
Windows machine, especially given a mixture of 
conda and pip envs, so no regression guard has been
added at this time

cc @rgommers @pv @peterjc123 @mattip @charris


tylerjereddy:
I don't think I can get away with this without at least a bit of shuffling in main repo `__init__.py`?

mattip:
Wouldn't it be easier to use `os.add_dll_directory(lib_path)` when possible?

pv:
The numpy.distutils upgrade should do the `os.add_dll_directory(lib_path)` bit.
It's maybe indeed better to move the distributor init import in the main repo after config import, similarly as it's in numpy.

pv:
I'd suggest moving `_distributor_init` import in the main repo: https://github.com/scipy/scipy/pull/12462

tylerjereddy:
@pv Ok, I've merged your main repo improvement, thanks.

Can you clarify--I just need to bump the submodule here now and remove the rest?

tylerjereddy:
Ok, I've adjusted this PR to bump `numpy-distutils` submodule and no other changes.

pv:
Ok, thanks. Yes, bumping the numpy-distutils here should ensure the correct DLL path is set with Python >= 3.8

pv:
Sorry, I'm wrong --- actually we also need to change `appveyor.yml` because currently we skip replacing numpy-distutils for Python >=3.8.

PR #91, opened 2020-07-19
BLD, MAINT: pin setuptools

tylerjereddy:
* attempt to avoid the SciPy main repo issue:
https://github.com/scipy/scipy/issues/12579

* normally we open wheels repo fixes to `master` branch,
but let's first just see if pinning `setuptools!=49.2.0`
can alleviate the SciPy `1.5.2` wheel build failures that
are observed on Linux only, as described above (a similar
strategy has recently been used in NumPy when these warnings
were observed there)

cc @rgommers @pv @mattip 

I'm not really sure why this is observed Linux-only, but the level of abstraction with shell scripts sourcing other shell functions through `multibuild` could mean that a  slightly different code path might be exercised for Linux `pip` commands or something like that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
         - NP_BUILD_DEP="numpy==1.14.5"
         - CYTHON_BUILD_DEP="Cython==0.29.18"
         - PYBIND11_BUILD_DEP="pybind11==2.4.3"
+        - SETUPTOOLS_BUILD_DEP="setuptools!=49.2.0"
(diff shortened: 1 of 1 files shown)


tylerjereddy:
That didn't help, I still see strange compile issues on Linux like `g++: error: unrecognized command line option ‘-std=c++14’`

tylerjereddy:
Older gcc toolchain on manylinux image having trouble with a new build flag or something? But why showing up just now?

PR #95, opened 2020-10-01
MAINT: expand ARM64 builds

tylerjereddy:
* expand the Travis CI ARM64 builds to include
Python 3.6 and Python 3.8 (since we need those wheels too), using separate
workspaces to pass data between stages for them

cc @janaknat @mattip @odidev  

Probably won't work on first try, but it looks like we [do have support](https://github.com/scipy/scipy/issues/11170#issuecomment-701674909) to release ARM64 wheels with `1.5.3`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,46 @@ jobs:
       after_success:
         - echo "This stage will not upload aarch64 wheel"
     - os: linux
+      arch: arm64-graviton2
+      dist: focal
+      virt: vm
+      group: edge
+      env:
+        - MB_PYTHON_VERSION=3.6
+        - PLAT=aarch64
+        - NP_BUILD_DEP=numpy==1.19.1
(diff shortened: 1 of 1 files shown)


mattip:
The build passed. The only thing maybe it would be nice to add is a [job name](https://docs.travis-ci.com/user/build-matrix/#job-names) so the [travis-ci.com summary](https://github.com/MacPython/scipy-wheels/pull/95/checks?check_run_id=1191187100) is a bit nicer.

tylerjereddy:
Ok, I've opened a separate issue for the "job names," which I  think extends beyond the ARM64 entries as well. I'll go ahead and self-merge with CI all green.

janaknat:
@tylerjereddy Will these changes land in 1.5.x and produce the aarch64 wheels?

janaknat:
Understood. Please let me know if I can help in any way.

PR #98, opened 2020-10-08
CI: turn on Python 3.9 in Travis CI

tylerjereddy:
* this will almost certainly fail for now,
but draft in changes for Python 3.9 entries
in Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,6 +93,27 @@ jobs:
       after_success:
         - echo "This stage will not upload aarch64 wheel"
     - os: linux
+      name: ARM64-Linux-Py39
+      arch: arm64-graviton2
+      dist: focal
+      virt: vm
+      group: edge
+      env:
+        - MB_PYTHON_VERSION=3.9
+        - PLAT=aarch64
+        - NP_BUILD_DEP=numpy==1.19.1
+        - CYTHON_BUILD_DEP="Cython"
+        - DOCKER_IMAGE=quay.io/pypa/manylinux2014_${PLAT}
+      script:
+        - echo "This stage will just build AArch64 wheel"
+      workspaces:
+        create:
+          name: ws4
+          paths:
+            - wheelhouse
+      after_success:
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Actually, not bad, I only see two Travis CI fails

tylerjereddy:
I restarted the two failing Python 3.9 jobs now that NumPy has wheels for 3.9.

The two original failures were in MacOS + 3.9 & Linux 32-bit + 3.9; fingers crossed.. if those builds succeed then we also have to manually restart the second stage I think, but one thing at a time..

tylerjereddy:
@mattip @charris Will a multibuild version bump help for MacOS here?

And for the 32-bit Linux failure--is there still a NumPy 32-bit wheel that might be missing for older manylinux or something like that? Not sure if that is a likely cause of an issue there.

charris:
> Will a multibuild version bump help for MacOS here?

Likely, NumPy is using the latest version and needed it for the Mac builds.

charris:
> And for the 32-bit Linux failure

Are you running the pre-releases? The 32 bit manylinux1 pre-releases are failing, but the manylinux2010 versions are good and current. That problem is supposed to get fixed by the SIMD folks.

tylerjereddy:
@charris NumPy is pinned at `1.19.1` for building and testing of the 32-bit Python 3.9 Linux build. Perhaps worth trying `1.19.3` for actual 3.9 wheels?

The test failures are really minor--e-16 floating point precision mismatch in single positions of arrays.

tylerjereddy:
Ok, looks like  bumping the NumPy version to `1.19.3` solved the 32-bit Linux Python 3.9 issue; now it is just the Linux ARM64 + Python 3.9 single test failure: https://github.com/scipy/scipy/issues/13019

tylerjereddy:
CI is all green and we'll want to backport this once the 1.5.x branch is ready so we can get Linux/Mac OS 3.9 wheels in 1.5.4 release. Merging it in.

PR #99, opened 2020-10-15
MAINT: ARM64 & related backports

tylerjereddy:
* try adding the ARM64 wheel builds to the `v1.5.x` wheels branch--an early sniff test to see if we run into issues with provision of ARM wheels for Linux on this older release branch

* point the builds to the tip of the maintenance branch in the main repo while I'm at it to get latest build of the branch

* the only notable backport I left out from `master` was a `multibuild` submodule bump; can do that if needed but being conservative initially

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,24 @@ jobs:
       # Exclude the default Python 3.5 build
       - python: 3.5
   include:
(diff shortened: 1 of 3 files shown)


tylerjereddy:
Looks good now

PR #104, opened 2020-11-16
MAINT: remove Python 3.6 wheels

tylerjereddy:
* remove Python 3.6 wheels following:
https://github.com/scipy/scipy/pull/13081


rgommers (approved):
This is definitely right, it just deletes jobs. So let's just merge as is, ignoring the other build issue due to `-c99` flag.

I'll send a follow-up PR for bumping the numpy version on the remaining jobs.

PR #110, opened 2021-01-18
BLD: update commit hashes to point at tip of scipy mckib2:gcc48to52

mckib2:
- see https://github.com/scipy/scipy/pull/13347
- testing to see if wheels using gcc5 need extra compiler/linker arguments as in https://github.com/pypa/auditwheel/issues/125
- the scipy branch currently does not use any special compiler/linker arguments -- this is a fact-finding mission

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ env:
     global:
         - REPO_DIR=scipy
         # Also see DAILY_COMMIT below
-        - BUILD_COMMIT=ab1c0907f
+        - BUILD_COMMIT=68eb089
         - PLAT=x86_64
         - CYTHON_BUILD_DEP="Cython==0.29.18"
         - PYBIND11_BUILD_DEP="pybind11==2.4.3"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,7 @@ environment:
       WHEELHOUSE_UPLOADER_SECRET:
         secure:
             jIyaD+VWmTlDGXThsKAkiLq8iljgYHiriq+kEUuW9tHj67R5BapLxLjbfco2nt8Y
-      BUILD_COMMIT: ab1c0907f
+      BUILD_COMMIT: 68eb089
       DAILY_COMMIT: master
 
   matrix:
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -2,7 +2,7 @@ env:
     global:
         - REPO_DIR=scipy
         # Also see DAILY_COMMIT below
-        - BUILD_COMMIT=ab1c0907f
+        - BUILD_COMMIT=68eb089
```
tylerjereddy, line 5: A few thoughts here:

1) Why not open the PR against `master`? Unless there's a special reason to do otherwise, we'd normally do that and backport from there if we really need a backport.
2) If you really want to build your specific feature branch, you can try setting `BUILD_COMMIT` to the name of your feature branch (if you open a PR against `master` you'll notice that `BUILD_COMMIT` points at `master`.
3) I don't see anything special about your feature branch at the moment in terms of things the wheels repo would care about--I assume your plan is to make meaningful changes in that feature branch in response to any issues that crop up here in the wheels repo.

mckib2: @tylerjereddy Thanks, I think with your feedback I understand what's going on a little bit better.
1. Yes, I can open a new PR against master
2. If I open a PR against master, does `DAILY_COMMIT` need to change to the branch name?  or do I just need to make `BUILD_COMMIT=mckib2:boost`?
3. I was pointing to the wrong branch --  I will need to point to the actual feature branch that we need testing for (https://github.com/scipy/scipy/pull/13328) and modify the scripts here to use gcc5 and then mirror those changes in mckib2:gcc48to52 if any are needed

I will open a new PR tomorrow, thanks!

tylerjereddy:
Just the build commit to temporarily point to your branch

rgommers:
@mckib2 in your new PR, you may want to delete `.travisci.yml` to avoid triggering TravisCI until other CI systems are happy - we're very short on TravisCI credits.

mckib2:
@rgommers Yes, I can do that.  I ran out of time today, hopefully tomorrow I can another PR open and try again.

PR #111, opened 2021-01-20
CI: bump gcc4.8 to gcc5

mckib2:
Another shot in the dark:
- try to point to branch that uses C++14 features (`mckib2:boost`)
- try to use GCC5
- disable TravisCI for now

After looking at this a little more, I find it to be a lot different than the scipy CI and I'm still trying to get a grasp on what needs to happen.  Any nudges in the right direction are appreciated.


azure-posix.yml
```diff
@@ -66,6 +70,9 @@ jobs:
 
           before_install
 
+          git remote add mckib2 https://github.com/mckib2/scipy
+          git fetch mckib2 boost
```
tylerjereddy, line 74: Here's a summary of what I *think* is happening here. These two `git` commands may get applied properly, but in the wrong repo--as it stands they might be applied to the `scipy-wheels` repo root. Then, the commands below, which come from the separate`multibuild` library, do work inside the `scipy` repo/dir proper and they have explicit `cd $REPO_DIR` style commands to do the work there (`scipy` is, for example, a submodule/subfolder of the wheels repo). So the desired remote/branch does not exist in the submodule/repo, only in the parent repo where we don't need it.

Perhaps something like this might suit:

```bash
pushd .
cd $REPO_DIR
git remote add mckib2 https://github.com/mckib2/scipy
git fetch mckib2 boost
popd

```

I can see the various `cd $REPO_DIR` commands in use in the `build_wheel` commands in `multibuild`, so I'd likely suggest trying to do the same around your commands here, perhaps with a debug `pwd` or `ls -tlrh` or similar thrown in for good measure to probe current location if you want.

mckib2: Alright, I threw a couple of things against the wall based on your feedback and it looks like one is sticking.  It seems like it pulled the remote references but didn't actually create a local copy of the branch, so explicitly creating the local branch worked.  Now to see about what it takes to get it to build...

mckib2: So some building now, some have lapack errors that I don't fully understand.  Is that a consequence of the gfortran vendored by multibuild?

rgommers: Copying the error, cause it's not showing up in all failed builds:
```
_________________________ TestF77Mismatch.test_lapack __________________________
[gw1] linux -- Python 3.7.6 /venv/bin/python
/venv/lib/python3.7/site-packages/scipy/linalg/tests/test_build.py:55: in test_lapack
    more information.""")
E   AssertionError: Both g77 and gfortran runtimes linked in scipy.linalg.flapack ! This is
E   likely to cause random crashes and wrong results. See numpy INSTALL.rst.txt for
E   more information.
        deps       = [b'libgfortran', b'libgfortran']
        f          = <scipy.linalg.tests.test_build.FindDependenciesLdd object at 0xe15f25cc>
        self       = <scipy.linalg.tests.test_build.TestF77Mismatch object at 0xe15f222c>
```
I think the `g77` part of the message may be a little misleading. I suspect that there's two different `libgfortran` versions detected by `ldd` or some such thing. Or, the new `gfortran` you added doesn't work for 32-bit (all relevant CI failures are 32-bit Linux) and there's a `g77` installed on the system already that gets picked up.

If you modify the test to display the `stdout` of running `ldd` as part of the test failure message, that would probably show what is going on. Doing that would be a useful change to make permanently.

mckib2: Starting to take another look at this as it's blocking https://github.com/scipy/scipy/pull/13347 and https://github.com/scipy/scipy/pull/13328

mckib2: Here's the ldd output for the _flapack shared library:
```
(venv) [root@efe5cafd1496 scipy]# ldd build/testenv/lib/python3.8/site-packages/scipy/linalg/_flapack.cpython-38-i386-linux-gnu.so 
	linux-gate.so.1 =>  (0xf7f33000)
	libopenblas.so.0 => /usr/local/lib/libopenblas.so.0 (0xf67c4000)
	libgfortran.so.5 => /usr/lib/libgfortran.so.5 (0xf65df000)
	libm.so.6 => /lib/libm.so.6 (0xf65b5000)
	libgcc_s.so.1 => /lib/libgcc_s.so.1 (0xf6596000)
	libc.so.6 => /lib/libc.so.6 (0xf63ff000)
	libpthread.so.0 => /lib/libpthread.so.0 (0xf63e4000)
	libgfortran.so.4 => /usr/lib/libgfortran.so.4 (0xf628d000)
	/lib/ld-linux.so.2 (0xf7f34000)
	libquadmath.so.0 => /usr/lib/libquadmath.so.0 (0xf61fc000)
	libz.so.1 => /lib/libz.so.1 (0xf61e8000)
```

Seems almost identical to https://github.com/scipy/scipy/issues/8855 (modulo gfortran versions)

h-vetinari:
@mckib2 
Are you still working (or willing to work) on this PR? If not, I'd be willing to give this a shot (since scipy 1.8 is not far)

mckib2:
> @mckib2 Are you still working (or willing to work) on this PR? If not, I'd be willing to give this a shot (since scipy 1.8 is not far) - I'd most likely work on top of your commits in a separate PR.

@h-vetinari No, I am not.  Thinking back, I think I got stalled on not being familiar with the scipy wheel generation infrastructure, please feel free to pick this up

h-vetinari:
> please feel free to pick this up

OK, thanks! Looking closer at the changes & commits, I think I'll try to avoid pointing to different repos, so I'll probably start from scratch after all.

Issue #114, opened 2021-02-26
Version mismatch causes pip warning and no install


djhoese:
In my CI build I have:

```
          python -m pip install \
          --extra-index-url https://pypi.anaconda.org/scipy-wheels-nightly/simple/ \
          --trusted-host pypi.anaconda.org \
          --no-deps --pre --upgrade \
          matplotlib \
          numpy \
          pandas \
          scipy;
```

which produces the following:

```
Looking in indexes: https://pypi.org/simple, https://pypi.anaconda.org/scipy-wheels-nightly/simple/
Requirement already satisfied: matplotlib in /usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages (3.3.4)
Collecting matplotlib
  Downloading matplotlib-3.4.0rc1-cp38-cp38-manylinux1_x86_64.whl (10.3 MB)
Requirement already satisfied: numpy in /usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages (1.20.1)
Collecting numpy
  Downloading https://pypi.anaconda.org/scipy-wheels-nightly/simple/numpy/1.21.0.dev0%2B821.g572d5a1da/numpy-1.21.0.dev0%2B821.g572d5a1da-cp38-cp38-manylinux2010_x86_64.whl (15.5 MB)
Requirement already satisfied: pandas in /usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages (1.2.2)
Collecting pandas
  Downloading https://pypi.anaconda.org/scipy-wheels-nightly/simple/pandas/1.3.0.dev0%2B849.gab687aec38/pandas-1.3.0.dev0%2B849.gab687aec38-cp38-cp38-manylinux1_x86_64.whl (19.1 MB)
Requirement already satisfied: scipy in /usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages (1.6.0)
Collecting scipy
  Downloading https://pypi.anaconda.org/scipy-wheels-nightly/simple/scipy/1.7.0.dev0%2B5e9eb01/scipy-1.7.0.dev0%2B20210110040759_5e9eb01-cp38-cp38-manylinux1_x86_64.whl (27.3 MB)
[...]

PR #118, opened 2021-04-17
Stop renaming the nightly wheels by adding date-time to them

rgommers:
Renaming is bad, it causes a mismatch between file name and `METADATA` inside the wheel. This is the cause of gh-114. https://github.com/scipy/scipy/pull/13882 should have removed the need for this renaming, it prepends a monotonously increasing number (commit count since previous branch point for a minor version).

Closes gh-114.


.travis.yml
```diff
@@ -1,3 +1,8 @@
+# Get a less shallow clone; needed to get the commit count since
+# the previous release branch point right.
+git:
+  depth: 3000
```
charris, line 4: You could use `false` to disable the depth restriction.

rgommers: Yeah I know, I chose not to because I only need the commits till the last branch point. Getting all 25,000 of them can be quite slow.

rgommers: 3,000 should be enough for any minor release period I think.

azure-posix.yml
```diff
@@ -114,7 +113,6 @@ jobs:
       - bash: |
           set -e
           if [ $ANACONDA_ORG == "scipy-wheels-nightly" ]; then
-            source extra_functions.sh
             for f in wheelhouse/*.whl; do rename_wheel $f; done
```
charris, line 116: The if statement should probably be deleted.

rgommers: Ah yes. Me and bash aren't friends:(

charris: I note there is also a `rename_wheel` in `.travis.yml`.

charris: Correction, only `for f in wheelhouse/*.whl; do rename_wheel $f; done` still needs to be removed. The if also encompasses the upload.

rgommers: It didn't encompass the upload, all removed now.

charris:
The `.travis.yml` script needs fixing even though the relevant path is not currently reached. The `manylinux2014_aarch64` (travis) and `win_amd64` (appveyor) development builds are uploading to the staging area, looks like it happened on the last PR merge to master as neither runs scheduled nightly builds. The logic could probably use some improvement so they upload to the nightly area.

rgommers:
Good catch, done. 

> The logic could probably use some improvement so they upload to the nightly area.

Indeed, but some other time - I've already invested way too much time in CI issues in the past week.

I think it could use a bigger overhaul actually, right now I just want to get it green and fix bug where nightlies are not installable.

charris:
OT: Is there a reason to keep using appveyor? NumPy does the windows builds on azure. It is also easy to add a cron job to travis, daily, weekly, and monthly are options.

rgommers:
TravisCI and Azure are happy, Appveyor is untouched in this PR - so let's call it good, in it goes.

`Successfully installed scipy-1.7.0.dev0+805.6a02659` shows it works as expected.

> OT: Is there a reason to keep using appveyor? NumPy does the windows builds on azure. 

Not really, just inertia.

PR #119, opened 2021-04-18
Add -m32 to the 32-bit Python on Windows builds

rgommers:
This change is copied from the Azure setup in the main SciPy repo. It should fix the Pythran build issue.


rgommers:
All Windows builds are green, finally. In it goes.

PR #124, opened 2021-08-13
WIP, MAINT: try 3.10rc wheels

tylerjereddy:
* mimic the approach used by NumPy wheels
in: https://github.com/MacPython/numpy-wheels/pull/129

* what could go wrong.. (skip Travis for now)

I'm guessing it is not going to be as easy as just copying what @mattip and @charris did but isn't going to stop me from trying :) Some of our SciPy wheels infra/base images normally used are a bit older perhaps too?

I think even if we do succeed I'd be hesitant to actually put the rc wheel on PyPI vs. just making it available on the anaconda testbed.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,6 +19,10 @@ jobs:
       name: linux
       vmImage: ubuntu-18.04
       matrix:
+        amd64-linux-py310:
+          AZURE_PYTHON_VERSION: "3.9"
+          MB_PYTHON_VERSION: "3.10"
+          DOCKER_TEST_IMAGE: "multibuild/focal_{PLAT}"
         amd64-linux-py39:
           MB_PYTHON_VERSION: "3.9"
         32bit-amd64-linux-py39:
--- a/azure-posix.yml
+++ b/azure-posix.yml
@@ -59,7 +59,7 @@ jobs:
           set -e
           echo building for commit "$BUILD_COMMIT"
           pip install --upgrade virtualenv wheel setuptools
-          BUILD_DEPENDS="wheel oldest-supported-numpy Cython>=0.29.18 pybind11>=2.4.3 pythran"
+          BUILD_DEPENDS="wheel oldest-supported-numpy Cython>=0.29.24 pybind11>=2.4.3 pythran"
 
           source multibuild/common_utils.sh
           source multibuild/travis_steps.sh


tylerjereddy:
oh, is `config_ilp64.sh` a NumPy specific thing? I may need to dig a little

charris:
Yep, `config_ilp64.sh` was needed to build 64 bit blas wheels. If you stick with 32 bit blas integers you won't need it.

tylerjereddy:
After removing that, `numpy.distutils.system_info.NotFoundError: No BLAS/LAPACK libraries found.`

tylerjereddy:
We are way behind NumPy in terms of the OpenBLAS version we are shipping with wheels, not sure if related.

mattip:
> We are way behind NumPy in terms of the OpenBLAS version 

You might want to sync `tools/openblas_support.py` with the one in NumPy

mattip:
Maybe connected to downloading a [manylinux1 openblas](https://dev.azure.com/numpy/numpy/_build/results?buildId=18982&view=logs&j=7eb27247-8768-54fc-61e3-aed398435c62&t=e3034194-08eb-5fd3-6e1a-a3b7a962520f&l=277) which later on cannot be linked since it is [missing some gfortran symbols](https://dev.azure.com/numpy/numpy/_build/results?buildId=18982&view=logs&j=7eb27247-8768-54fc-61e3-aed398435c62&t=e3034194-08eb-5fd3-6e1a-a3b7a962520f&l=402)

mattip:
xref scipy/scipy#14316

tylerjereddy:
Makes sense, we got blocked by that linear algebra testing issue for 4-5 versions of OpenBLAS I think, so all the attempts to bump failed. Looks like someone is chipping away at the issue a little at least.

charris:
NumPy will only build manylinux2014 wheels for 1.22. I'm also going to release manylinux2014 3.10 wheels for 1.21.2 so more people can start testing.

charris:
You also need cython >= 0.29.22, but might as well use the latest version 0.29.24.

tylerjereddy:
Ah, I guess I had assumed I needed manylinux 2014, guess I can try removing that at least.

mattip:
manylinux 2014 is OK. There are two sides to moving off manylinux1 and manylinux2010:
- Will people be able to use `pip` to install the manylinux2014 wheel (yes, since python3.10 comes with pip-21.1.3 which supports manyinux2014)?
- does the glibc in the oldeset distro supported conform to manylinux2014/glibc2.17 (yes, since CenOS6 (manylinux2010, glibc 2.15) is [EOL](https://github.com/mayeut/pep600_compliance#acceptable-distros-to-build-wheels)

tylerjereddy:
I'll close this--I think @tupui is going to try adding 3.10 now at @rgommers suggestion--for some reason I thought we might have some issues that we'd been avoiding with manylinux version bump but hopefully I"m wrong

PR #125, opened 2021-08-20
Fixes for latest commit of scipy and add universal2 wheels

isuruf:
cc @judahrand

Wait for
- [x] https://github.com/scipy/scipy/pull/14631
- [x] https://github.com/matthew-brett/trusty/pull/15
- [x] https://github.com/scipy/scipy/pull/14633

universal2 support was added in https://github.com/scipy/scipy/pull/14316

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -8,6 +8,8 @@ function build_wheel {
         unset FFLAGS
         export LDFLAGS="-shared -Wl,-strip-all"
         build_libs $PLAT
+        # Remove once https://github.com/scipy/scipy/pull/14631 lands
+        export CFLAGS="$CFLAGS -D_ISOC99_SOURCE=1"
         # Work round build dependencies spec in pyproject.toml
         build_bdist_wheel $@
     else
--- a/config.sh
+++ b/config.sh
@@ -16,12 +16,29 @@ function build_wheel {
     fi
 }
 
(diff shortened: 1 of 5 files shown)


isuruf:
This is finally ready for a review @rgommers.

config.sh
```diff
@@ -4,34 +4,24 @@
 source gfortran-install/gfortran_utils.sh
 
 function build_wheel {
+    export FFLAGS="$FFLAGS -fPIC"
     if [ -z "$IS_OSX" ]; then
-        unset FFLAGS
-        export LDFLAGS="-shared -Wl,-strip-all"
```
rgommers, line 9: Just checking: this can be removed because it's done as a post-processing step by multibuild (https://github.com/matthew-brett/multibuild/blob/cdffc52846355707f389808414cf8dfe31289328/configure_build.sh#L44)? I'm not quite sure why in that code macOS is special-cased and doesn't have `STRIP_FLAGS`.

isuruf: Previously, because of numpy's LDFLAGS behaviour (overwriting instead of appending), this was needed, but not anymore.
`-shared` is not needed anymore due to that and strip flags are set by multibuild anyway.

rgommers (approved):
Looks great, let's give this a go! Thanks @isuruf and @judahrand!

PR #127, opened 2021-09-30
MAINT: add Python 3.10 wheels

tupui:
Build wheels for Python 3.10.

xref https://github.com/scipy/scipy/issues/14567 and #124

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ jobs:
       virt: vm
       group: edge
       env:
-        - MB_PYTHON_VERSION=3.10.0
+        - MB_PYTHON_VERSION=3.10
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
         - DOCKER_IMAGE=quay.io/pypa/manylinux2014_${PLAT}
@@ -104,7 +104,7 @@ jobs:
       virt: vm
       group: edge
       env:
-        - MB_PYTHON_VERSION=3.10.0
+        - MB_PYTHON_VERSION=3.10
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
         - DOCKER_TEST_IMAGE=multibuild/focal_{PLAT}
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,9 +75,46 @@ jobs:
             - wheelhouse
       after_success:
         - echo "This stage will not upload aarch64 wheel"
+    - os: linux
+      name: ARM64-Linux-Py310
+      arch: arm64-graviton2
+      dist: focal
+      virt: vm
+      group: edge
+      env:
+        - MB_PYTHON_VERSION=3.10
+        - PLAT=aarch64
+        - CYTHON_BUILD_DEP="Cython"
+        - DOCKER_IMAGE=quay.io/pypa/manylinux2014_${PLAT}
+        - MB_ML_VER=2014
+      script:
+        - echo "This stage will just build AArch64 wheel"
+      workspaces:
+        create:
+          name: ws4
+          paths:
+            - wheelhouse
+      after_success:
+        - echo "This stage will not upload aarch64 wheel"
 
     - stage: Test wheel
       arch: arm64-graviton2
+      name: ARM64-Linux-Py310
+      dist: focal
+      virt: vm
+      group: edge
+      env:
+        - MB_PYTHON_VERSION=3.10
+        - PLAT=aarch64
+        - CYTHON_BUILD_DEP="Cython"
+        - DOCKER_TEST_IMAGE=multibuild/xenial_arm64v8
+        - MB_ML_VER=2014
+      workspaces:
+        use: ws4
+      install:
+        - echo "This stage will test and upload the AArch64 wheel"
+
+    - arch: arm64-graviton2
       name: ARM64-Linux-Py39
       dist: focal
       virt: vm
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ jobs:
       virt: vm
       group: edge
       env:
-        - MB_PYTHON_VERSION=3.10
+        - MB_PYTHON_VERSION=3.10.0
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
         - DOCKER_IMAGE=quay.io/pypa/manylinux2014_${PLAT}
@@ -104,7 +104,7 @@ jobs:
       virt: vm
       group: edge
       env:
-        - MB_PYTHON_VERSION=3.10
+        - MB_PYTHON_VERSION=3.10.0
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
         - DOCKER_TEST_IMAGE=multibuild/xenial_arm64v8
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,7 +107,7 @@ jobs:
         - MB_PYTHON_VERSION=3.10.0
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
-        - DOCKER_TEST_IMAGE=multibuild/xenial_arm64v8
+        - DOCKER_TEST_IMAGE=multibuild/focal_{PLAT}
         - MB_ML_VER=2014
       workspaces:
         use: ws4
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
         # Also see DAILY_COMMIT below
         - BUILD_COMMIT=master
         - PLAT=x86_64
-        - CYTHON_BUILD_DEP="Cython==0.29.18"
+        - CYTHON_BUILD_DEP="Cython==0.29.24"
         - PYTHRAN_BUILD_DEP="pythran"
         - PYBIND11_BUILD_DEP="pybind11==2.4.3"
         - UNICODE_WIDTH=32
--- a/.travis.yml
+++ b/.travis.yml
@@ -123,7 +123,7 @@ jobs:
         - MB_PYTHON_VERSION=3.9
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
-        - DOCKER_TEST_IMAGE=multibuild/xenial_arm64v8
+        - DOCKER_TEST_IMAGE=multibuild/focal_{PLAT}
         - MB_ML_VER=2014
       workspaces:
         use: ws4
@@ -138,7 +138,7 @@ jobs:
         - MB_PYTHON_VERSION=3.8
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
-        - DOCKER_TEST_IMAGE=multibuild/xenial_arm64v8
+        - DOCKER_TEST_IMAGE=multibuild/focal_{PLAT}
         - MB_ML_VER=2014
       workspaces:
         use: ws3
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,12 +91,11 @@ jobs:
         - echo "This stage will just build AArch64 wheel"
       workspaces:
         create:
-          name: ws4
+          name: ws5
           paths:
             - wheelhouse
       after_success:
         - echo "This stage will not upload aarch64 wheel"
-
     - stage: Test wheel
       arch: arm64-graviton2
       name: ARM64-Linux-Py310
@@ -110,10 +109,9 @@ jobs:
         - DOCKER_TEST_IMAGE=multibuild/focal_{PLAT}
         - MB_ML_VER=2014
       workspaces:
-        use: ws4
+        use: ws5
       install:
         - echo "This stage will test and upload the AArch64 wheel"
-
     - arch: arm64-graviton2
       name: ARM64-Linux-Py39
       dist: focal
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,6 +27,20 @@ environment:
       DAILY_COMMIT: master
 
   matrix:
+    - PYTHON: C:\Python310
+      PYTHON_VERSION: 3.10
+      PYTHON_ARCH: 32
+      NUMPY_BUILD_DEP: numpy==1.19.3
+      NUMPY_TEST_DEP: numpy==1.19.3
+      CYTHON_BUILD_DEP: Cython
+
+    - PYTHON: C:\Python310-x64
+      PYTHON_VERSION: 3.10
+      PYTHON_ARCH: 64
+      NUMPY_BUILD_DEP: numpy==1.19.3
+      NUMPY_TEST_DEP: numpy==1.19.3
+      CYTHON_BUILD_DEP: Cython
+
     - PYTHON: C:\Python39
       PYTHON_VERSION: 3.9
       PYTHON_ARCH: 32
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ clone_depth: 50
 # No reason for us to restrict the number concurrent jobs
 max_jobs: 100
 
-image: Visual Studio 2017
+#image: Visual Studio 2017
 
 cache:
   - '%LOCALAPPDATA%\pip\Cache'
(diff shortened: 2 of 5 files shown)


tupui:
@tylerjereddy for azure, do I need to do `AZURE_PYTHON_VERSION: "3.9"` as you had in your PR?

tylerjereddy:
For ARM64 Python 3.10 issue, looks like we need to bump `pip` version up slightly to pick up this patch: https://github.com/pypa/pip/pull/10252 -- it may be worth trying to bump the version of the `multibuild` submodule to the latest version to see if that helps.


> for azure, do I need to do AZURE_PYTHON_VERSION: "3.9" as you had in your PR?

Not sure, I guess NumPy was/is doing that because `3.10` wasn't available for the base CI image yet, but `3.10` release candidate was available in the docker images used to build the wheels?

charris:
> do I need to do AZURE_PYTHON_VERSION: "3.9"

Python is used in two places: to run the docker image and inside the docker image. Azure doesn't have 3.10 yet, so 3.9 is used to run the docker image, which in turn uses 3.10.0 internally. For 3.9, it was about three weeks after release before azure had it.

tupui:
> > do I need to do AZURE_PYTHON_VERSION: "3.9"
> 
> Python is used in two places: to run the docker image and inside the docker image. Azure doesn't have 3.10 yet, so 3.9 is used to run the docker image, which in turn uses 3.10.0 internally. For 3.9, it was about three weeks after release before azure had it.

ok so I need to do this right
```
AZURE_PYTHON_VERSION: "3.9"
MB_PYTHON_VERSION: "3.10.0"
DOCKER_TEST_IMAGE: "multibuild/focal_{PLAT}"  # this as well?
```

azure-pipelines.yml
```diff
@@ -19,6 +19,11 @@ jobs:
       name: linux
       vmImage: ubuntu-18.04
       matrix:
+        amd64-linux-py10:
+          MB_PYTHON_VERSION: "3.10.0"
+        32bit-amd64-linux-py10:
+          MB_PYTHON_VERSION: "3.10.0"
+          PLAT: "i686"
```
tupui, line 26:
```suggestion
        amd64-linux-py10:
          AZURE_PYTHON_VERSION: "3.9"
          MB_PYTHON_VERSION: "3.10.0"
        32bit-amd64-linux-py10:
          AZURE_PYTHON_VERSION: "3.9"
          MB_PYTHON_VERSION: "3.10.0"
          PLAT: "i686"
```

azure-pipelines.yml
```diff
@@ -49,3 +54,10 @@ jobs:
           MB_PYTHON_VERSION: "3.9"
           MB_PYTHON_OSX_VER: "10.9"
           PLAT: universal2
+        osx-Py310:
+          MB_PYTHON_VERSION: "3.10.0"
+          MB_PYTHON_OSX_VER: "10.9"
+        osx-Py310-universal2:
+          MB_PYTHON_VERSION: "3.10.0"
+          MB_PYTHON_OSX_VER: "10.9"
+          PLAT: universal2
```
tupui, line 63:
```suggestion
        osx-Py310:
          AZURE_PYTHON_VERSION: "3.9"
          MB_PYTHON_VERSION: "3.10.0"
          MB_PYTHON_OSX_VER: "10.9"
        osx-Py310-universal2:
          AZURE_PYTHON_VERSION: "3.9"
          MB_PYTHON_VERSION: "3.10.0"
          MB_PYTHON_OSX_VER: "10.9"
          PLAT: universal2
```

tupui (commented):
@charris please commit if this is correct

tupui:
> For ARM64 Python 3.10 issue, looks like we need to bump `pip` version up slightly to pick up this patch: [pypa/pip#10252](https://github.com/pypa/pip/pull/10252) -- it may be worth trying to bump the version of the `multibuild` submodule to the latest version to see if that helps.

Sorry @tylerjereddy I don't get how to update the submodule nor the pip version...

tylerjereddy:
I usually do something like `git submodule update --init` then manually `cd multibuild` and pull in the latest `devel` branch from https://github.com/matthew-brett/multibuild ; then back up to main repo and commit the change to where the submodule is pointing

.travis.yml
```diff
@@ -75,9 +75,46 @@ jobs:
...
       arch: arm64-graviton2
+      name: ARM64-Linux-Py310
+      dist: focal
+      virt: vm
+      group: edge
+      env:
+        - MB_PYTHON_VERSION=3.10.0
+        - PLAT=aarch64
+        - CYTHON_BUILD_DEP="Cython"
+        - DOCKER_TEST_IMAGE=multibuild/xenial_arm64v8
```
hugovk, line 110: I think you need focal here? See https://github.com/matthew-brett/multibuild/issues/421#issuecomment-898354903

azure-pipelines.yml
charris:
```
AZURE_PYTHON_VERSION: "3.9"
MB_PYTHON_VERSION: "3.10.0"
DOCKER_TEST_IMAGE: "multibuild/focal_{PLAT}"  # this as well?
```
Focal is needed to get python 3.10 but only has 64 bit versions, so no 32 bit wheels. I used `"3.10"` for the version, guess we will see if `"3.10.0"` works as well :) NumPy is going to drop 32 bit manylinux wheels for 3.10 and only keep them for Windows.

.travis.yml
tupui: Thanks!

azure-pipelines.yml
tupui: Ok so I should remove the manylinux 32 bit for 3.10?

tupui: Indeed, 3.10.0 fails for azure. I dropped the 32 manylinux as well since you do this in NumPy.

charris:
NumPy has also moved to manylinux2014 everywhere. See https://github.com/MacPython/numpy-wheels/blob/main/azure-pipelines.yml.

charris:
Note that 3.10 isn't yet available for OS X.

tupui:
> NumPy has also moved to manylinux2014 everywhere. See https://github.com/MacPython/numpy-wheels/blob/main/azure-pipelines.yml.
> 
> EDIT: Also https://github.com/MacPython/numpy-wheels/blob/main/.travis.yml

Ok I think I fixed it.



> Note that 3.10 isn't yet available for OS X.

You are talking about this one?
```
        osx-Py310-universal2:
          AZURE_PYTHON_VERSION: "3.9"
          MB_PYTHON_VERSION: "3.10"
          MB_PYTHON_OSX_VER: "10.9"
          PLAT: universal2
```
Because the not `universal` seemed to go.

charris:
> Heh, the manylinux version bump is what I was trying to avoid

There were enough changes in the 1.22.x wheels that a NEP might have been useful.

- Change to ILP64 OpenBLAS
- Dropping manylinux1 and manylinux2010
- Dropping 32 bit Python 3.10 manylinux wheels
- Dropping Python 3.7

The changes were all discussed in triage meetings or on the mailing list, but that probably didn't help downstream projects.

rgommers:
I think we should follow the NumPy choices, except let's not do ILP64 OpenBLAS. Happy to drop manylinux2010.

tupui:
@rgommers If I understood everything, should be it.

Travis is failing for `ARM64-Linux-Py39` with `ERROR: no supported wheels found`. I am not sure to what it refers to. Only difference is the Cython version but it should be [there](https://pypi.org/project/Cython/0.29.24/#files) no?

rgommers:
The log is, as usual with pip, remarkably uninformative. Comparing with the successful build for Python 3.8, this is what it should have looked like:
```
...
You should consider upgrading via the '/venv/bin/python -m pip install --upgrade pip' command.
Looking in links: https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com
Processing ./wheelhouse/scipy-1.8.0.dev0+1759.3159565-cp38-cp38-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
Requirement already satisfied: numpy>=1.17.3 in /venv/lib/python3.8/site-packages (from scipy==1.8.0.dev0+1759.3159565) (1.19.2)
Installing collected packages: scipy
Successfully installed scipy-1.8.0.dev0+1759.3159565
```
It's unclear why the Rackspace URL didn't work - it's hard to access without the password/token. I've restarted the failed job, let's see if it happens again.

mattip:
I am late to the party, but what is this still using a rackspace URL? That died a while ago, we moved to anaconda.org

rgommers:
There's Rackspace links all over the place for multibuild. Maybe multiple submodules need a bump in this repo?

mattip:
It is hard to see what is going on [in the log](https://app.travis-ci.com/github/MacPython/scipy-wheels/jobs/540717058) since it is not echoing the commands inside the docker. The rackspace url is a false flag, it should not be needed. Could you add `set -ex` to the `install_run` command in `config.sh`? That should tell us which wheel it is failing to find. Strange that this is in the python3.9 run, which worked in the past.

mattip:
OK, so this happened `python /io/multibuild/supported_wheels.py /io/wheelhouse/scipy-1.8.0.dev0+1759.3159565-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl` which means somehow the **test** step of 3.9 got the wheel from the **build** step with a 3.10 wheel

mattip:
I think the additions to the travis.yml are mixed a bit. There is a build for python3.9, then a build for python3.10, then a test for python 3.10, then a test for python3.9. Maybe this should be refactored to use a matrix, or at least to build/test, build/test

tupui:
Seems like it was it 😄 Hopefully when appveyor decides to run, it's all good.

charris:
> I think the additions to the travis.yml are mixed a bit.

The travis scripts look overly complicated. It isn't clear to me why `graviton2` and `edge` are needed, I suspect they are historical artifacts from the time ARM64 builds suddenly started failing. The staging also complicated things. But those cleanups should probably happen in another PR.

.travis.yml
```diff
@@ -86,7 +121,7 @@ jobs:
         - MB_PYTHON_VERSION=3.9
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
-        - DOCKER_TEST_IMAGE=multibuild/xenial_arm64v8
```
charris, line 89: These two don't need the update to focal for testing. The reason to use different docker images for building and testing is to test on old Ubuntu versions. That couldn't be done for 3.10 because, IIRC, deadsnakes didn't provide 3.10 for xenial. @mattip Do you recall the reasons? I note that deadsnakes was recently updated.

charris:
The Appveyor builds are all failing
```
cipy\special\_cosine.c(101): error C2059: syntax error: 'bad suffix on number'
```
Indeed, looks like it has been failing for some time: https://anaconda.org/scipy-wheels-nightly/scipy/files?version=1.8.0.dev0%2B1753.a063cf3. The nightly files could use a deletion cleanup, but I don't have the permissions.

tylerjereddy:
I'll try to do some purging there this weekend, thanks Chuck et al. for all the feedback here (and Pamphile for driving this forward).

We may need to open a main repo issue for the `special` problem for it to get some attention?

charris:
> we should migrate the TravisCI jobs to GitHub Actions

Appveyor also needs a migration, it takes three hours to complete, and that is when it fails without testing.

charris:
The appveyor problem seems to be the hex float
```
#define M_PI64 0x1.921fb54442d18p+1
```
a c99 thing apparently not supported in the MSVC version used for the wheel builds.

rgommers:
> The appveyor problem seems to be the hex float
> 
> ```
> #define M_PI64 0x1.921fb54442d18p+1
> ```
> 
> a c99 thing apparently not supported in the MSVC version used for the wheel build

Argh, continuing headache with that one. I hope @WarrenWeckesser knows the right thingy to replace it with.

charris:
> Argh, continuing headache with that one.

Lot of those in NumPy main too, it no longer compiles with visual studio 2015 (the appveyor default).  The version can be changed by adding `image: Visual Studio 2019` to the `appveyor.yml`, see https://www.appveyor.com/docs/build-environment/

rgommers:
Ah okay, that's a better solution. The version matching Python 3.7 is `Visual Studio 2017 version 15.7`, so that's an improvement.

tupui:
> > Argh, continuing headache with that one.
> 
> Lot of those in NumPy main too, it no longer compiles with visual studio 2015 (the appveyor default). The version can be changed by adding `image: Visual Studio 2017` to the `appveyor.yml`, see https://www.appveyor.com/docs/build-environment/

@charris looks like this is causing an error. Or I am not using this correctly.

charris:
Hmm, looks like that example wasn't quite correct, see https://www.appveyor.com/docs/appveyor-yml/. Maybe the capitalization needs to be `image: Visual Studio 2017`?

tupui:
> Hmm, looks like that example wasn't quite correct, see https://www.appveyor.com/docs/appveyor-yml/. Maybe the capitalization needs to be `image: Visual Studio 2017`?

mmm no I had this right (just checked and I even have PyCharm complaining if I change the casing).

Not sure about the output, but seems like `Microsoft SDKs` is not present in `C:\Program Files` then.
```
dir "C:/Program Files/Microsoft SDKs/Windows"
120 Volume in drive C has no label.
121 Volume Serial Number is 1CB8-D4AA
122
123 Directory of C:\Program Files\Microsoft SDKs
124
125File Not Found
```

charris:
I'm kind of lost then. That scipy needs fortran complicates things.

tupui:
Ok fixed the path thing, let's see now where it fails... 😅

charris:
Hmm, looks like MSVC 2017 compiler is 14.1. I'm looking at  https://wiki.python.org/moin/WindowsCompilers, which isn't very explicit.

tupui:
Then this should be fine??
```
 - copy "C:\Program Files (x86)\Microsoft Visual Studio\VC\redist\x86\Microsoft.VC141CRT\msvcp141dll" "build\lib.win32-%PYTHON_VERSION%\scipy\.libs\"
  - copy "C:\Program Files (x86)\Microsoft Visual Studio\VC\redist\x64\Microsoft.VC141CRT\msvcp141dll" "build\lib.win-amd64-%PYTHON_VERSION%\scipy\.libs\"
```

But before this, it's not building NumPy's wheel now.

charris:
I wouldn't worry about Python 3.10 on windows yet, there are no NumPy 3.10 wheels for that platform yet.

tupui:
Oh ok but then it's ready because Travis and Azure are passing. I will comment the changes on appveyor and it should be green.

charris:
Appveyor was already failing because of the hex float, so I think the MSVC update is still needed. But is could be another PR.

appveyor.yml
```diff
@@ -146,8 +163,8 @@ build_script:
   # The version should match the compiler version, see https://wiki.python.org/moin/WindowsCompilers
   - mkdir "build\lib.win32-%PYTHON_VERSION%\scipy\.libs"
   - mkdir "build\lib.win-amd64-%PYTHON_VERSION%\scipy\.libs"
-  - copy "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\redist\x86\Microsoft.VC140.CRT\msvcp140.dll" "build\lib.win32-%PYTHON_VERSION%\scipy\.libs\"
-  - copy "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\redist\x64\Microsoft.VC140.CRT\msvcp140.dll" "build\lib.win-amd64-%PYTHON_VERSION%\scipy\.libs\"
+  - copy "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\redist\x86\Microsoft.VC140CRT\msvcp14.0dll" "build\lib.win32-%PYTHON_VERSION%\scipy\.libs\"
```
charris, line 166: `VC140.CRT\msvcp140.dll`, not `VC140CRT\msvcp14.0dll`.  But still needs to upgrade to visual studio 2017.

charris:
Seems to be working.

charris:
Thanks @tupui .

tupui:
> Thanks @tupui .

Well thanks to you @charris!

tylerjereddy:
I went ahead and purged out about 6 months of old SciPy nightly (weekly) wheels uploads to the anaconda site--there shouldn't be anything more than 7-8 days old there now. Thanks for the reminder.

PR #130, opened 2021-10-02
MAINT: Use manylinux2014 for everything manylinux.

charris:


charris:
Scipy compiled with Python 3.7, 3.8 on manylinux2014 results in  [6 test errors]( https://tinyurl.com/yjaktcox) when tested on xenial,

tylerjereddy:
>  do we want to drop all 32 bit manylinux wheels?

Maybe good for Ralf to confirm, but I suspect if NumPy has dropped all 32-bit scenarios except Windows (where 32-bit still allegedly represents a substantial portion of use cases), we might be "ok" with following suit.

I might be hesitant to make that a backported change (i.e., my impression from Matthew Brett was that we may not want to change our offerings on a given minor release series like 1.7.2 vs 1.7.1).

charris:
Update, manylinux2010 does work for the 32 bit manylinux wheels.

rgommers:
> Maybe good for Ralf to confirm, but I suspect if NumPy has dropped all 32-bit scenarios except Windows (where 32-bit still allegedly represents a substantial portion of use cases), we might be "ok" with following suit.

Agreed, we should do the same as NumPy here.

> I might be hesitant to make that a backported change (i.e., my impression from Matthew Brett was that we may not want to change our offerings on a given minor release series like 1.7.2 vs 1.7.1).

It'd be better not to indeed, if we can avoid it (and that should be possible). We should only add new wheels in a bugfix release, not change the manylinux version or drop a wheel.

For 1.7.2, things that worked before should still work, and all we want to do there is add new Python 3.10 jobs, right?

tylerjereddy:
Sounds right, so I guess we can drop the 32-bit manylinux wheels on `master` of wheels repo

tylerjereddy (approved):
The CI is green and I think we're on the same page here.

We can probably follow-up by just deleting the 32-bit Linux entries in another PR if that is the direction NumPy is going, but seems like we don't want to fall too far behind with manylinux versions compared to NumPy/ecosystem so those changes make sense to me.

tylerjereddy:
Thanks Chuck and Ralf

Issue #129, opened 2021-10-02
travis and appveyor upload nightly builds to the wrong repository.


charris:
The builds are going to "multibuild-wheels-staging" instead of "scipy-wheels-nightly".

rgommers:
Fixed in the main repo, closing.

PR #133, opened 2021-10-03
MAINT: Update test docker version

charris:
- Change test docker image to xenial except for Python 3.10.
- Change logic so both pushes and merges to master upload to nightly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,7 +121,7 @@ jobs:
         - MB_PYTHON_VERSION=3.9
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
-        - DOCKER_TEST_IMAGE=multibuild/focal_{PLAT}
+        - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
         - MB_ML_VER=2014
       workspaces:
         use: ws4
@@ -136,7 +136,7 @@ jobs:
         - MB_PYTHON_VERSION=3.8
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
-        - DOCKER_TEST_IMAGE=multibuild/focal_{PLAT}
+        - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
         - MB_ML_VER=2014
       workspaces:
         use: ws3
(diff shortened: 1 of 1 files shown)


charris:
Travis won't run for me anymore.  I suspect they may have marked me as spamming the tests :))

mattip:
It is not an issue with credits, the MacPython group still has about 30% of the credits left

.travis.yml
```diff
@@ -136,7 +136,7 @@ jobs:
         - MB_PYTHON_VERSION=3.8
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
-        - DOCKER_TEST_IMAGE=multibuild/focal_{PLAT}
+        - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
```
tylerjereddy, line 139: Is the motivation here to test with "oldest supported" image/dependencies where possible, instead of bleeding edge?

.travis.yml
```diff
@@ -167,24 +167,23 @@ script:
...
-    # trigger an upload to the shared ecosystem
-    # infrastructure at: https://anaconda.org/scipy-wheels-nightly
-    # for cron jobs only (restricted to master branch once
-    # per week)
-    # SCIPY_WHEELS_NIGHTLY is a secret token
-    # used in Travis CI config, originally
-    # generated at anaconda.org for scipy-wheels-nightly
-    - if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
+    # upload wheels for pushes and merges.
+    - if [ "$TRAVIS_BRANCH" == "master" ]; then
```
tylerjereddy, line 171: Do we really want to have new wheels in the nightly repo for a reason other than the weekly crons? I thought the weekly cadence was intentional--since the wheels repo can go quiet for > 1 month at a time, if we really wanted to consistently provide i.e., daily updates to the bleeding edge "nightly" wheels wouldn't we want that to be reflected in the frequency of the crons?

.travis.yml
```diff
@@ -167,24 +167,23 @@ script:
...
-          anaconda -t ${SCIPY_WHEELS_NIGHTLY} upload --force -u ${ANACONDA_ORG} ${TRAVIS_BUILD_DIR}/wheelhouse/*.whl;
-      fi
-    # for merges (push events) we use the staging area instead;
-    # SCIPY_STAGING_UPLOAD_TOKEN is a secret token used in Travis
-    # CI config, originally generated at anaconda.org for
-    # multibuild-wheels-staging
-    - if [ "$TRAVIS_EVENT_TYPE" == "push" ]; then
+          TOKEN=${SCIPY_WHEELS_NIGHTLY};
+      else
+          # for merges (push events) we use the staging area instead;
```
tylerjereddy, line 179: With `TRAVIS_EVENT_TYPE` filter removed, that means we do not block the `pull_request` event for backport activity in wheels repo.

I'd have to double check to be sure, but the secrets may be restricted to branches originating in the repo proper instead of from forks, which may possibly cause Travis CI to error (and perhaps even explain the lack of a CI matrix entry for Travis here?).

More generally though, I'm at least equally confused about the new settings vs. the old ones. I just checked, and I don't actually see Travis CI cron jobs running for SciPy wheels at the moment/last 6 months, perhaps because Ralf and I might have decided to purge the cron when the job credits stuff happened? (there is no cron job in our wheels repo settings for Travis at the moment).

tylerjereddy:
I've tried to clarify what is going on  with uploads in gh-129--if there's a compelling reason to switch our philosophy to match NumPy we could do that, but I think some of my comments there may clear some things up (and also clarify that some cron jobs simply don't exist and/or are disabled).

rgommers:
No longer needed, so closing. Thanks all!

PR #135, opened 2021-10-26
WIP, MAINT: backport 3.10 support

tylerjereddy:
* try to backport some of the pertinent
changes for Python 3.10 support to the wheels
repo `v1.7.x` branch, which needs to continue
supporting Python 3.7

* perhaps we can do away with some of the manylinux
bumps--not sure on those--things are going to change
a fair bit anyway with a huge bump in OpenBLAS version
as well; perhaps the bumps could be confined to just 3.10
using newer manylinux??

* point `BUILD_COMMIT` at the tip of the target maintenance
branch to get more useful feedback on where we stand
for the 3.10 wheel builds

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/scipy/scipy.git
 [submodule "multibuild"]
 	path = multibuild
-	url = https://github.com/matthew-brett/multibuild.git
+	url = https://github.com/multi-build/multibuild.git
 [submodule "gfortran-install"]
 	path = gfortran-install
 	url = https://github.com/MacPython/gfortran-install.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
     global:
         - REPO_DIR=scipy
         # Also see DAILY_COMMIT below
-        - BUILD_COMMIT=1bbc2b4
+        - BUILD_COMMIT=5f587c6
         - PLAT=x86_64
         - CYTHON_BUILD_DEP="Cython==0.29.24"
         - PYTHRAN_BUILD_DEP="pythran"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,11 @@ env:
     global:
         - REPO_DIR=scipy
         # Also see DAILY_COMMIT below
-        - BUILD_COMMIT=47bb6feb
+        - BUILD_COMMIT=1bbc2b4
         - PLAT=x86_64
-        - CYTHON_BUILD_DEP="Cython==0.29.18"
+        - CYTHON_BUILD_DEP="Cython==0.29.24"
         - PYTHRAN_BUILD_DEP="pythran"
-        - PYBIND11_BUILD_DEP="pybind11==2.4.3"
+        - PYBIND11_BUILD_DEP="pybind11>=2.4.3"
         - UNICODE_WIDTH=32
         - MANYLINUX_URL="https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com"
         - WHEELHOUSE_UPLOADER_USERNAME=travis-worker
@@ -44,6 +44,7 @@ jobs:
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
         - DOCKER_IMAGE=quay.io/pypa/manylinux2014_${PLAT}
+        - MB_ML_VER=2014
       script:
         - echo "This stage will just build AArch64 wheel"
       workspaces:
@@ -64,6 +65,7 @@ jobs:
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
         - DOCKER_IMAGE=quay.io/pypa/manylinux2014_${PLAT}
+        - MB_ML_VER=2014
       script:
         - echo "This stage will just build AArch64 wheel"
       workspaces:
@@ -84,6 +86,7 @@ jobs:
         - PLAT=aarch64
         - CYTHON_BUILD_DEP="Cython"
         - DOCKER_IMAGE=quay.io/pypa/manylinux2014_${PLAT}
+        - MB_ML_VER=2014
       script:
         - echo "This stage will just build AArch64 wheel"
       workspaces:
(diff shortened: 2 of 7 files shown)


rgommers:
> I'm also wondering if we really can justify `universal2` wheels for MacOS Python 3.8 and 3.9? The next planned point release is for `3.10` support, not M1 support yet (those jobs do pass CI though).

Yes, we should not ship `universal2` wheels unless we know the segfaults are fixed (which they are not at this point).

rgommers:
> Python 3.10 on MacOS has an error like the following, and perhaps a newer MacOS image may be needed there or something like that:

This doesn't look familiar to me (but I haven't followed the recent 3.10 work closely).

> Python `3.10` elsewhere simply isn't considered a supported version for SciPy until we backport the `pyproject.toml`/metadata support for it in the main repo.

This is done now in the main repo.

tylerjereddy:
Awesome, thanks I'll probably merge that in the main repo then revise this PR based on comments and to point to the latest version of the maintenance branch hash so we have an updated view on the situation for 1.7.2.

tylerjereddy:
Ok, latest update pushed--quoting from the commit message:

```
    * we're not ready to ship `universal2` wheels for Python 3.8
    and 3.9, so remove those backport additions; for Python 3.10
    `universal2` is apparently our only option (based on the comments
    I've seen from Chuck at least?)
    
    * still skipping travis CI for now to save credits
    
    * point `BUILD_COMMIT` at latest version of the maintenance
    branch to get updated feedback on our status here

```

tylerjereddy:
@charris @matthew-brett do I understand correctly that for MacOS X Python 3.10 `universal2` is our only option, even though I only want to ship Intel Mac support for this point release? I'm basing this just on having read some comments/release information related to NumPy I think.

charris:
>  MacOS X Python 3.10 universal2 is our only option

That was my understanding. Python 3.10 only supports universal2 installers, but I'm not sure what that means.  However, I did notice that thin `macosx_11` arm64 wheels were also produced  by universal2, so I suspect Intel wheels could be produced, but manylinux won't do it that I know of.  Note that multibuild sets the mac build version to 11+ for 3.10 even though the universal2 wheel is marked 10_9. I can't find much documentation on this stuff. If scipy fails on M1 silicon, it may be a problem...

rgommers:
> That was my understanding. Python 3.10 only supports universal2 installers, but I'm not sure what that means. 

That cannot be correct. Maybe you mean that Python.org offers only a universal2 installer? If so, I don't see how that is relevant. We should produce wheels that make sense, so also thin `x86_64` and `arm64` wheels in the future. And for now only a `x86_64` one, unless there's some pragmatic reason that that is difficult. 

If it is difficult, then shipping `universal2` wheels for Python 3.10 alone seems okay, with a note that there may be issues left for arm64 users. It's still better than letting all 3.10 users build from source.

> However, I did notice that thin macosx_11 arm64 wheels were also produced by universal2, so I suspect Intel wheels could be produced, but multibuild won't do it that I know of. Note that multibuild sets the mac build version to 11+ for 3.10 even though the universal2 wheel is marked 10_9. I can't find much documentation on this stuff. If scipy fails on M1 silicon, it may be a problem...

Ah. That seems wrong. But well, multibuild is a bit of a pain to update and we really need to ship `1.7.2`, so let's just do what is easiest.

charris:
> so let's just do what is easiest.

If the Intel wheel problem gets solved, those wheels can always be uploaded later. Universal 3.10 wheels were OK for NumPy, so I stopped trying to get the Intel builds working. There might be a solution there, I just didn't find it in the short time I devoted to it.

rgommers:
> Universal 3.10 wheels were OK for NumPy, so I stopped trying to get the Intel builds working. There might be a solution there, I just didn't find it in the short time I devoted to it.

For NumPy it's fine right now, because there's nothing really wrong with numpy on arm64, tests pass modulo a few warnings. SciPy still has crashes in several modules.

charris:
> there's nothing really wrong with numpy on arm64

Yep. If you want, I could do more work on getting the Intel build working. NumPy takes a lot less time to build and test than SciPy does.

rgommers:
> EDIT: One of the reasons I stopped trying is that multibuild might need fixing, and there was no easy way to do that while working in numpy-wheels, the commits need to be in the upstream multibuild devel branch.

Yeah I'm inclined to spend time on moving to cibuildwheel instead. Long term, multibuild updates are lost dev time.

tylerjereddy:
Alright, well let me see if I hit a wall with at least trying to build a regular x86_64 3.10 wheel for Mac, probably in a separate trimmed down debug PR and/or with some local testing..

tylerjereddy:
Something, I assume `multibuild`, is really insisting on the dual `-arch arm64 -arch x86_64` with `clang` for Python `3.10`. I tried gently circumventing with a multibuild hack but now I'm just trying to delete everything mac-related that uses `arm64` in a debug `multibuild` branch for sanity..

mattip:
Maybe worth pinging @matthew-brett on how to make x86_64 wheels for macOS

tylerjereddy:
Ok, it is not looking like multibuild is injecting it directly on my test branch (I purged every opportunity to add it from multibuild). I've read about Xcode 12/the system Python itself automatically injecting the flag, which is what I feared, so will see if one of the described workarounds is viable.. `export ARCHFLAGS="-arch x86_64"` seems the best bet if I can find just the rigth way to add it (hopefully no prepend vs. append flag nightmares).

tylerjereddy:
The Python 3.10 wheel build appears to succeed in gh-136 with x86_64 arch forced, but there are quite a few test failures in the next stage, especially related to `float128` and `complex256` array results dtype comparisons in `sparse`. I'll relay similar info in the main repo to get a few more eyes on it.

charris:
> especially related to float128 and complex256

Interesting. Those are both double on M1 and numpy ran into problems with c++ there, but that shouldn't be the case on Intel.

tylerjereddy:
Following various upstream updates the following changes have been pushed in here (from the commit message):

```
    * switch to `x86_64` thin wheel builds
    for Mac OS Python `3.10` following upstream
    changes/releases in multibuild/NumPy that
    support this

    * point to the new `multibuild` repo and sync
    to latest `devel` (for thin MacOS Python 3.10
    x86_64 thin wheel support, etc.)

    * backport the full set of `master` branch changes
    to `config.sh` based on feedback from Isuru
```

tupui:
For the failures on macOS, I've seen the same in my PR to update our CI https://github.com/scipy/scipy/pull/14832

tylerjereddy:
@tupui @charris @rgommers I suspect if we're going to use `oldest-support-numpy` we're either going to have update that repo and do another release there because we need absolute bleeding edge NumPy wheels for MacOS 3.10; alternatively, at least for the backport PR here, might we consider just putting the actual NumPy version we need for that one MacOS Python 3.10 job?

At the moment those failures are identical to before, and almost certainly the result of not using the latest NumPy with the quad/double fixes.

tupui:
> @tupui @charris @rgommers I suspect if we're going to use `oldest-support-numpy` we're either going to have update that repo and do another release there because we need absolute bleeding edge NumPy wheels for MacOS 3.10; alternatively, at least for the backport PR here, might we consider just putting the actual NumPy version we need for that one MacOS Python 3.10 job?

Ralf just updated it so it serves 1.21.4. So maybe just relaunch the CI?

tylerjereddy:
Ok, let's just re-flush the CI then. I'd try to restart just that one job, but let's just check across the board..

tylerjereddy:
So close, just this issue left it seems: https://github.com/scipy/scipy/issues/14985

tylerjereddy:
Not sure if it is good fortune or the `import` added to `run_scipy_tests.py` allowing appveyor to pass this time, but let's not dwell on that and move forward with the release process..

PR #136, opened 2021-10-30
DEBUG only: MacOS x86_64 Python 3.10 wheel attempts

tylerjereddy:
- debug splitoff of gh-135 where the first commit cuts out most of the CI except MacOS Azure
- the second commit is for iterating on MacOS x86_64 Python 3.10 SciPy wheel build


tylerjereddy:
Looks like we are using `-arch arm64 -arch x86_64` in the Python `3.10` build, perhaps via the multibuild machinery, resulting in errors like this:

```
/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/machine/_mcontext.h:31:2: error: architecture not supported
#error architecture not supported
```

Nearly identical errors have been reported in many other places like here: https://github.com/facebookresearch/detectron2/issues/2288#issuecomment-730308073

Suggested solutions include `export ARCHFLAGS="-arch x86_64"`.

tylerjereddy:
The multibuild README has a section on MacOS binary behavior with a provision on overriding `wrap_wheel_builder()` if you don't want to go with one of the available options. I've attempted to provide a separate control flow that proceeds with a native intel build without fusing or ARM-related stuff.

PR #138, opened 2021-11-05
MAINT: multibuild new repo update

tylerjereddy:
* point multibuild at the new repo location
and use latest `devel`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/scipy/scipy.git
 [submodule "multibuild"]
 	path = multibuild
-	url = https://github.com/matthew-brett/multibuild.git
+	url = https://github.com/multi-build/multibuild.git
 [submodule "gfortran-install"]
 	path = gfortran-install
 	url = https://github.com/MacPython/gfortran-install.git


charris:
You should also be able to build thin macosx wheels for Python 3.10.

tylerjereddy:
that's true, and indeed I'm attempting that for the backport in gh-135. I wasn't actually sure if we want to expand our `master` branch offerings just yet or if we might still aim for `universal2` on `master`; certainly for that backport PR I've set to thin x86_64 Mac OS Python 3.10 only

rgommers:
We should also do thin wheels on master - thin wheels are the preferred format, universal2 is kind of a hack that we could leave out at some point.

charris:
I'm going to speculate that the problems come from `#include <numpy/ndarrayobject.h>`.

tylerjereddy:
The Python 3.8 Windows permission failure in `C:\Python38-x64\lib\site-packages\scipy\integrate\tests\test__quad_vec.py:127: in test_quad_vec_pool_args` is super annoying--that's affecting the backport/1.7.2 release as well..

tylerjereddy:
ok, well we need to point to the new `multibuild` repo location moving forward anyway, so I'll put this in

There is indeed some more stuff to do with thin wheels here and in the release branch..

PR #139, opened 2021-11-08
BUG: Includes msvcp140.dll in python 3.10 build on windows

thomasjpfan:
Fixes https://github.com/scipy/scipy/issues/14998

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ environment:
 
   matrix:
     - PYTHON: C:\Python310-x64
-      PYTHON_VERSION: 3.10.0
+      PYTHON_VERSION: '3.10'
       PYTHON_ARCH: 64
       NUMPY_BUILD_DEP: oldest-supported-numpy
       NUMPY_TEST_DEP: oldest-supported-numpy


tylerjereddy:
I'll need to backport the fix as well if this is what we want. Is there a simple explanation for why that one-liner allows the dll to get included? A version comparison or path construction that happens somewhere or something like that?

charris:
Ugh, I actually noticed that, but it seemed to work and I forgot about it. I've been taught this lesson many times...

thomasjpfan:
>  Is there a simple explanation for why that one-liner allows the dll to get included?

`msvcp140.dll` is always copied into `build\lib.win-amd64-%PYTHON_VERSION%`. On the `master` branch, we have `PYTHON_VERSION=3.10.0` which expanded into `build\lib.win-amd64-3.10.0`. This is not the same directory of the wheel: `build\lib.win-amd64-3.10`.

tylerjereddy (approved):
Windows Python 3.10 CI now shows the apparently desired:

`copying build\lib.win-amd64-3.10\scipy\.libs\msvcp140.dll -> build\bdist.win-amd64\wheel\.\scipy\.libs
`

It sounds like Ralf may be "ok" with trying a build tag increment, and indeed time is tight if we can save from a new release today, so I'll proceed with merging this, backporting, and then see if I can do that increment properly with changes in the wheels repo only.

tylerjereddy:
the other CI failures are not related and/or do not affect the maintenance branch for 1.7.x anyway

tylerjereddy:
thanks Thomas

PR #142, opened 2021-11-20
WIP: add MacOS arm64 thin wheel builds

tylerjereddy:
* try adding Python `3.10` arm64 MacOS thin wheel build for `v1.7.x` branch, with latest `BUILD_COMMIT` on the release branch--more changes still coming in main repo, so this is just an early check on what might go wrong beyond what we already know (usually there are some surprises we want to know about early on..)
* temporarily reduce the CI matrix to save resources while debugging

* I think @rgommers will help a bit with the `_distributor_init.py` reqs (https://github.com/scipy/scipy/issues/13409#issuecomment-974681045) tomorrow, or at least give me some pointers if I can't find an example of using it selectively on a wheel build with some histroy/grep checking

Questions:
1. @isuruf this is gong to cross-compile, but probably not run the tests? (I assume the current state of M1 "emulation" and trying to run the huge SciPy test suite would be prohibitively costly/complex, etc.)
2. @rgommers are we going to restrict the thin arm64 MacOS wheels to Python `3.10`, or do you think we do/should have a viable route to go back a few more versions (for example, this branch supports `3.7`--something tells me that older version + arm64 M1 would be a nightmare).
3. Am I correct in understanding that we need `MB_PYTHON_OSX_VER: "10.12"` for the new thin wheels because supporting older versions can literally cause a computer to restart with a kernel panic?


tylerjereddy:
Current raw build log: https://dev.azure.com/numpy/27346c6a-2774-4eac-bf85-e068127c0ccc/_apis/build/builds/21286/logs/12

- `gfortran: error: unrecognized command line option ‘-arch’`
- `build/lib.macosx-11.0-arm64-3.10/scipy/integrate/_quadpack.cpython-310-darwin.so" failed with exit status 1`

rgommers:
> @isuruf this is gong to cross-compile, but probably not run the tests? (I assume the current state of M1 "emulation" and trying to run the huge SciPy test suite would be prohibitively costly/complex, etc.)

Indeed, we can't run the tests. We just upload the wheel to the Anaconda staging bucket, and test locally (I have the hardware).

> @rgommers are we going to restrict the thin arm64 MacOS wheels to Python 3.10, or do you think we do/should have a viable route to go back a few more versions (for example, this branch supports 3.7--something tells me that older version + arm64 M1 would be a nightmare).

With the multibuild setup we should aim for the same Python version support as NumPy 1.21.4 - so 3.8-3.10 (3.7 won't work).

> Am I correct in understanding that we need `MB_PYTHON_OSX_VER: "10.12"` for the new thin wheels because supporting older versions can literally cause a computer to restart with a kernel panic?

No, that's an old version. macOS versioning is `10.x ... 10.16, 11.0-11.6, 12.0`. Haven't tested, but I think you want `12.0` here.

The current pre-release builds are for `master` and macOS `11.0` - see https://anaconda.org/scipy-wheels-nightly/scipy/files. I suggest getting them updated first, then the backport should be a lot easier.

tylerjereddy:
Ah, `master` of wheels is a bit messier, not even using the right `multibuild` at the moment. I may mess around here a bit more before giving up..

tylerjereddy:
hrm, maybe need a few adjustments in places: `Python download failed! Check python-3.8.10-macosx12.0.pkg exists on the server.'`

tylerjereddy:
Ok, updated to reflect backports from:

- gh-143
- gh-144
- gh-145
- `gfortran-install` sync to `master` hash

tylerjereddy:
Ok, the CI is passing now, and the 3 new thin wheel builds match what I saw on `master`. Appveyor is queued into oblivion, but not adjusted here substantially anyway..

PR #143, opened 2021-11-21
MAINT: set `OPENBLAS_THREAD_TIMEOUT` on arm64 macOS to avoid performance issues

rgommers:
Addresses at least part of https://github.com/scipy/scipy/issues/15050,
TBD if that closes it or if we need to rebuild OpenBLAS instead later.
For now (and for the 1.7.3 release) this is the safest and quickest
improvement.

### DIFF
--- a/_distributor_init.py
+++ b/_distributor_init.py
@@ -10,11 +10,14 @@ can safely replace this file with your own version.
 """
 
 import os
+import sys
+import platform
 
-# on Windows SciPy loads important DLLs
-# and the code below aims to alleviate issues with DLL
-# path resolution portability with an absolute path DLL load
+
 if os.name == 'nt':
+    # on Windows SciPy loads important DLLs
+    # and the code below aims to alleviate issues with DLL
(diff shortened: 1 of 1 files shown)


tylerjereddy (approved):
Awesome, thanks Ralf. I'll backport it as well

I can't actually test it myself, but the CI is green and it looks like exactly what I would have done.

ogrisel:
@rgommers @tylerjereddy It seems that this `_distributor_init.py` file is not shipped in the macos arm64 wheels which were generated after the merge of this PR:

https://anaconda.org/scipy-wheels-nightly/scipy/files

ogrisel:
It's only included in the windows wheels in the appveyor config:

https://github.com/MacPython/scipy-wheels/blob/05bdc404f7daef5799c3b23c66e069ae4fd13c57/appveyor.yml#L142-L147

rgommers:
Thanks @ogrisel, I'll fix that up now.

PR #144, opened 2021-11-22
MAINT: add thin wheels, some fixes

tylerjereddy:
* forward-port multiprocessing addition to `run_scipy_tests.py`, which
was helpful in preventing sporadic Appveyor/Windows test failures on the
`v1.7.x` branch

* forward-port Python `3.10` x86_64 thin wheel support

* add Python 3.8 to 3.10 MacOS arm64 thin wheel builds for
MacOS 12; I'll be fairly surprised if the issues for Python < 3.10
download with this super-new OS version observed in the backport PR
are not observed on `master` (need multibuild change or don't actually
need to bump the version that high to restrict to MacOS 12 only
usage of the wheel?)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -55,6 +55,10 @@ jobs:
           MB_PYTHON_VERSION: "3.8"
           MB_PYTHON_OSX_VER: "10.9"
           PLAT: universal2
+        osx-Py38_arm64:
+          MB_PYTHON_VERSION: "3.8"
+          MB_PYTHON_OSX_VER: "12.0"
+          PLAT: arm64
         osx-Py39:
           MB_PYTHON_VERSION: "3.9"
           MB_PYTHON_OSX_VER: "10.9"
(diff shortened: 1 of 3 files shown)


tylerjereddy:
I've asked for some advice in the cross-linked `multibuild` issue.

tylerjereddy:
Ok, one getting closer now.. just the `3.8` and `3.9` MacOS thin arm64 wheel entries failed. I'm pretty sure that without `multibuild` changes I'll have to use `MB_PYTHON_OSX_VER: 10.9` for those two entries, so I'll try that now (`11.0` is "ok" for Python 3.10).

tylerjereddy:
Ok, everything looks good in CI now--the 3 new arm64 thin wheels are built and their testing is obviously skipped. The new x86_64 Python 3.10 thin wheel is built & tested.

tylerjereddy:
merging, and proceed with wheels repo backporting effort

rgommers:
There's still one more thing to do: rename the wheels from `11_0_arm64` to `12_0_arm64`. I think that should be a separate post-processing step in CI here, because it's specific to SciPy.

For reference, original string is defined at https://github.com/multi-build/multibuild/blob/devel/osx_utils.sh#L108-L109

PR #147, opened 2021-11-23
MAINT: rename MacOS arm64 wheels

tylerjereddy:
* automatically rename MacOS arm64 thin wheel files to
use version `12_0` instead of `11_0`; the verbose file
move should print out the change for us if successful

### DIFF
--- a/azure-posix.yml
+++ b/azure-posix.yml
@@ -83,6 +83,12 @@ jobs:
         condition: ne(variables['PLAT'], 'arm64')
 
       - bash: |
+          set -xe
+          for file in `find ./wheelhouse -type f -name '*11_0_arm64*.whl'`; do mv -v "$file" "${file/11_0_arm64/12_0_arm64}"; done
+        displayName: "Rename MacOS arm64 wheels for version 12.0 minimum"
+        condition: eq(variables['PLAT'], 'arm64')
+
+      - bash: |
           echo "##vso[task.prependpath]$CONDA/bin"
           sudo chown -R $USER $CONDA
         displayName: Add conda to PATH


azure-posix.yml
```diff
@@ -82,6 +82,12 @@ jobs:
         displayName: Install wheel and test
         condition: ne(variables['PLAT'], 'arm64')
 
+      - bash: |
+          set -xe
+          for file in `find ./wheelhouse -type f -name '*11_0_arm64*.whl'`; do mv -v "$file" "${file/11_0_arm64/12_0_arm64}"; done
+        displayName: "Rename MacOS arm64 wheels for version 12.0 minimum"
+        condition: eq(variables['PLAT'], 'arm64')
```
tylerjereddy, line 89: this is slight overkill since there's only one wheel per matrix entry, but it does seem to work locally and the paradigm was conveniently available from a stackoverflow snippet.. let's see what CI says..

rgommers (commented):
While we're at it: the `universal2` wheels should also be renamed to `12_0` (now `scipy-1.8.0.dev0+2085.d18b2a4-cp310-cp310-macosx_10_9_universal2.whl`).

tylerjereddy:
That's true, though CI is passing and we don't have universal2 wheels on the release branch so maybe we can merge for now and open an issue.

rgommers:
Sure, that's fine with me - in it goes. Thanks Tyler!

rgommers:
For future archeologists: this was because of the kernel panic on macOS 11: https://github.com/scipy/scipy/issues/14688#issuecomment-971211057

PR #150, opened 2021-11-27
WIP, MAINT: MacOS wheels 12.0 target

### LINKED ISSUE #148: universal2 wheels need to be renamed to `...12_0...`
This is a follow up to gh-147 that is needed. `arm64` were renamed to ensure they can only be installed on macOS >= 12.0, however `universal2` wheels need to be renamed in the same way.

tylerjereddy:
Fixes #148

* `universal2` wheels should now target MacOS 12.0+
using the approached described upstream:
https://github.com/multi-build/multibuild/issues/444#issuecomment-977573836

* fix the renaming of MacOS arm64 thin wheels to use
the same approach instead of my previous manual bash
commands

* note that, confusingly, `MACOSX_DEPLOYMENT_TARGET` is
also used in `env_vars.sh` and in `azure-posix.yml`, so
we should double check the CI logs before merging to make
sure that `pip` named the wheels correctly/we get the desired
overrides for the cases touched here

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/scipy/scipy.git
 [submodule "multibuild"]
 	path = multibuild
-	url = https://github.com/multi-build/multibuild.git
+	url = https://github.com/tylerjereddy/multibuild.git
 [submodule "gfortran-install"]
 	path = gfortran-install
 	url = https://github.com/MacPython/gfortran-install.git
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -55,10 +55,12 @@ jobs:
           MB_PYTHON_VERSION: "3.8"
           MB_PYTHON_OSX_VER: "10.9"
           PLAT: universal2
+          MACOSX_DEPLOYMENT_TARGET: "12.0"
         osx-Py38_arm64:
           MB_PYTHON_VERSION: "3.8"
           MB_PYTHON_OSX_VER: "10.9"
           PLAT: arm64
+          MACOSX_DEPLOYMENT_TARGET: "12.0"
         osx-Py39:
           MB_PYTHON_VERSION: "3.9"
           MB_PYTHON_OSX_VER: "10.9"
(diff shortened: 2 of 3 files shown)


tylerjereddy:
No current intention to backport this one, I'll work on a separate change for the distribution file problem that we'd likely want to backport for a `1.7.4.`

tylerjereddy:
CI says:

> Overwriting readonly variable 'MACOSX_DEPLOYMENT_TARGET' is not permitted

Not sure we should really have that set read-only anyway.. this seems like a clear use case where overriding is desired.

tylerjereddy:
Ok, trying to make that var writeable now..

rgommers:
This didn't work yet, wheels are still named like: `scipy-1.8.0.dev0+2099.3f05efb-cp310-cp310-macosx_11_0_arm64.whl`

isuruf:
https://github.com/multi-build/multibuild/blob/041625d47880e11ae3d2333874027e749b35fa72/osx_utils.sh#L489 should be fixed to not overwrite

isuruf:
I mean not overwrite if the value does not start with `10`

tylerjereddy:
Ah ok, so we need an upstream fix first. That's ok, I can try to help there if needed.

tylerjereddy:
I took a shot at the multibuild adjustment in the cross-linked PR, still untested though.

tylerjereddy:
I pushed a commit to point this PR at least temporarily at my `multibuild` feature branch on my fork, until this PR is merged: https://github.com/multi-build/multibuild/pull/446

It should at least tell us if the thin and universal2 wheels get properly renamed with that upstream change, rather than losing an overnight testing period.

tylerjereddy:
CI is passing but this looks wrong, i.e., 

- `/Users/runner/work/1/s/wheelhouse/scipy-1.9.0.dev0+1042.d634f17-cp39-cp39-macosx_11_0_arm64.whl`
- `./wheelhouse/scipy-1.9.0.dev0+1042.d634f17-cp310-cp310-macosx_10_9_universal2.whl`

I think at this stage I'm going to have to pursue more manual wheel renaming until things are actually working upstream with `MACOSX_DEPLOYMENT_TARGET`..

rgommers:
No longer needed as we don't do `universal2` builds - closing.

PR #145, opened 2021-11-22
BUG: now actually include `_distributor_init.py` in macOS arm64 wheel

rgommers:
This is a follow-up to gh-143

PR #151, opened 2021-11-28
MAINT, CI: check _distributor_init.py contents

tylerjereddy:
* following on from continued problems reported here:
https://github.com/scipy/scipy/issues/15050

* draft in code to check that MacOS wheels get
the same `_distributor_init.py` as the one present
in the wheels repo, rather than the generic nearly-empty
copy actually present in the `1.7.3` release

* we've had two pull requests about this and it slipped through into
a release despite 3 sets of eyes so need to be careful here;
I'd like to see the test script added here failing for MacOS in CI
like it does locally, before attempting a fix

* this is likely still WIP, since relative paths/directories
can be confusing to predict in the wheels/multibuild control
flow, which also flushes in and out of containers I think, etc.
(this probably contributed to the original problem)

* the script did seem to work in a quick local check though, complaining
about our new MacOS file inclusion, but happy with the long-standing
windows wheel inclusion for dirs containing wheels for each OS type:

`python verify_init.py /Users/treddy/rough_work/scipy/release_173/release/installers/windows`
`python verify_init.py /Users/treddy/rough_work/scipy/release_173/release/installers/linux`
`python verify_init.py /Users/treddy/rough_work/scipy/release_173/release/installers/mac`

```
  File "verify_init.py", line 40, in <module>
    check_for_dist_init(input_dir=input_dir)
  File "verify_init.py", line 34, in check_for_dist_init
    raise ValueError(f"Contents of _distributor_init.py incorrect for
{wheel_file}")
ValueError: Contents of _distributor_init.py incorrect for
/Users/treddy/rough_work/scipy/release_173/release/installers/mac/scipy-1.7.3-cp39-cp39-macosx_10_9_x86_64.whl
```

* we could go a step farther here and error out on Linux if the contents of `_distributor_init.py` *did* match the copy in the wheels repo, since they should not for that platform, though my primary focus is to be strict on the platforms that *do* need the adjustments (MacOS/Windows)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -198,6 +198,8 @@ test_script:
   - python ..\run_scipy_tests.py %TEST_MODE% -- -n6 --junitxml=%cd%\junit-results.xml -rfEX
 
 after_test:
+  - ls ..\scipy\dist
+  - python verify_init.py ..\scipy\dist
   # Upload test results to Appveyor
   - ps: |
       If (Test-Path .\junit-results.xml) {
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -198,8 +198,10 @@ test_script:
   - python ..\run_scipy_tests.py %TEST_MODE% -- -n6 --junitxml=%cd%\junit-results.xml -rfEX
 
 after_test:
-  - ls ..\scipy\dist
-  - python verify_init.py ..\scipy\dist
+  - cd ..
+  - ls scipy\dist
+  - python verify_init.py scipy\dist
+  - cd tmp_test
   # Upload test results to Appveyor
   - ps: |
       If (Test-Path .\junit-results.xml) {
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -143,7 +143,7 @@ build_script:
       $cwd = Get-Location
       ls $cwd
       rm -Force $cwd/scipy/scipy/_distributor_init.py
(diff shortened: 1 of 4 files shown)


rgommers:
Failures look as expected here. I'll push the fix for inclusion of the file on macOS and see that this turns green.

rgommers:
The bug was in gh-145, the fix was missed a second `/scipy`.

tylerjereddy:
The appveyor fails probably mean I still don't fully grasp the working directory situation there, I suppose we could just delete the check in appveyor for now if we only want to address the immediate issue

rgommers:
> The appveyor fails probably mean I still don't fully grasp the working directory situation there, I suppose we could just delete the check in appveyor for now if we only want to address the immediate issue

That looks like the right order of doing things to me.

rgommers:
I did look at it, and I don't see an obvious mistake in what you did in `appveyor.yml`.

isuruf:
https://github.com/MacPython/scipy-wheels/blob/master/appveyor.yml#L146 needs to be a copy instead of a move

tylerjereddy:
good catch, I've pushed the appveyor adjustment in

tylerjereddy:
Ok, CI is perfect now, I'll proceed with merge and backport.

PR #157, opened 2021-12-19
Fix universal2 wheel name and remove redundant builds

isuruf:
cc @tylerjereddy

### DIFF
--- a/azure-posix.yml
+++ b/azure-posix.yml
@@ -86,11 +86,11 @@ jobs:
           set -xe
           for file in `find ./wheelhouse -type f -name '*11_0_arm64*.whl'`; do mv -v "$file" "${file/11_0_arm64/12_0_arm64}"; done
         displayName: "Rename MacOS arm64 wheels for version 12.0 minimum"
-        condition: eq(variables['PLAT'], 'arm64')
+        condition: or(eq(variables['PLAT'], 'arm64'), eq(variables['PLAT'], 'universal2'))
 
       - bash: |
           set -xe
-          for file in `find ./wheelhouse -type f -name '*10_9_universal2.whl'`; do mv -v "$file" "${file/10_9_universal2/12_0_universal2}"; done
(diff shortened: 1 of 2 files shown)


azure-posix.yml
```diff
@@ -86,11 +86,11 @@ jobs:
           set -xe
           for file in `find ./wheelhouse -type f -name '*11_0_arm64*.whl'`; do mv -v "$file" "${file/11_0_arm64/12_0_arm64}"; done
         displayName: "Rename MacOS arm64 wheels for version 12.0 minimum"
-        condition: eq(variables['PLAT'], 'arm64')
+        condition: or(eq(variables['PLAT'], 'arm64'), eq(variables['PLAT'], 'universal2'))
 
       - bash: |
           set -xe
-          for file in `find ./wheelhouse -type f -name '*10_9_universal2.whl'`; do mv -v "$file" "${file/10_9_universal2/12_0_universal2}"; done
+          for file in `find ./wheelhouse -type f -name '*10_9_universal2.whl'`; do mv -v "$file" "${file/macosx_10_9_universal2/macosx_12_0_arm64.macosx_10_9_x86_64}"; done
```
tylerjereddy, line 93: Are you changing the name of the wheel to no longer include `universal2`? I dunno, I was just hoping to minimize the number of changes that get backported, even if the risk is slim. I guess you may be pretty confident it will do the right thing..

Still, I'm not sure we are particularly keen on retaining `universal2` long-term anyway..

isuruf: Yes. It's still a universal2 wheel, but this is to workaround a pip limitation.

tylerjereddy: Ok, I'll let @rgommers decide between this and gh-156. I'd just as soon dump `universal2` stuff sooner than later, but I can see both sides...

tylerjereddy:
Ok, I'll trust that you're close to 100 % confident that this won't introduce new problems compared to the simple solution in gh-156. Merging, thanks.

Issue #159, opened 2021-12-20
BUG: all 64-bit Windows wheel builds are failing


tylerjereddy:
- `1.8.0rc2` appveyor build matrix: https://ci.appveyor.com/project/scipy/scipy-wheels/builds/41957637
- from a `git diff ...` of the `1.8.0rc1` vs. `1.8.0rc2` Python 3.9 64-bit build logs, I find the following notable differences:
  - setuptools `59.5.0` vs. setuptools `60.0.1`
  - cython 0.29.25 vs. cython 0.29.26
  - pythran 0.10.0 vs 0.11.0
- I suspect we're getting prohibited versions of those deps because `python setup.py bdist_wheel` is used instead of i.e., `pip wheel` or `pip install` type commands that go through `pyproject.toml`? the appveyor config probably needs an update---if I can cheat and pin the deps another way for now I may do that, but complete resolution of this issue probably means modernizing the build approach (and we plan on replacing appveyor eventually anyway I think...)

rgommers:
Definitely pin `setuptools` - `60.0.0` re-enabled its vendored `distutils` copy, and that will almost certainly break things.

rgommers:
Fixed in our `cibuildwheel` builds, so closing.

PR #160, opened 2021-12-20
MAINT: pin deps for 1.8.x

tylerjereddy:
Proposed temporary workaround for gh-159 on `v1.8.x` branch

* at least temporarily pin the appveyor
Windows build dependencies that jumped
between 1.8.0rc1 and 1.8.0rc2, likely
causing the latter to fail

* this is not likely something that should
be forward ported; instead, we should likely
start leveraging newer PEP-compliant install
commands on Windows/Appveyor instead
of `python setup.py..`


rgommers (approved):
LGTM, let's see if CI turns green.

tylerjereddy:
CI is green, including appveyor, so merged.

I'll resume release management duties tomorrow morning, when the new rc2 wheels should be built off this merge event.

Issue #162, opened 2022-01-21
BUG: cl.exe used instead of clang-cl.exe for Windows pythran building


tylerjereddy:
Ralf and I have seen this before in the main repo (in fact, we temporarily disabled Pythran on Azure Windows there IIRC). 

Now I'm seeing the same thing for the `1.8.0rc3` wheel builds: https://ci.appveyor.com/project/scipy/scipy-wheels/builds/42265533 (Looking at the `3.10` builds specifically in analysis below).

The compile line switched from `clang-cl.exe` to `C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\HostX86\x64\cl.exe`, and the latter is causing a failed Pythran-related compile in the new RC build. (MSVC version paths do look the same before/after)

Almost all dependency versions appear to be identical by looking at the logs. `oldest-supported-numpy-0.15` was incremented up from `0.14` in the previously succeeding log, but that seems like stretch to cause the issue (NumPy version looks the same). We also already pinned the "usual suspects" between rc1 and rc2 so not surprising we're stable on the versions of most things.

What to do about it?

One (short-term) thought, browsing through mostly-unrelated `pythran` issue tracker discussions, is to identify the path to `clang-cl.exe` on the Appveyor image, then hard-code the path like here in `.pythranrc` https://github.com/serge-sans-paille/pythran/issues/1921#issuecomment-986586953:

```
[compiler]
CC=C:\PROGRA~2\MIB055~1\2019\BUILDT~1\VC\Tools\Llvm\x64\bin\clang-cl.exe
CXX=C:\PROGRA~2\MIB055~1\2019\BUILDT~1\VC\Tools\Llvm\x64\bin\clang-cl.exe
```

rgommers:
Hard coding the path to `clang-cl.exe` is worth a try. If that fails, then let's build without Pythran.

tylerjereddy:
not a problem anymore after gh-167

PR #163, opened 2022-01-21
WIP, BUG: debug work for clang-cl

tylerjereddy:
I've temporarily pared down the CI to a single Appveyor run while I look into gh-162.


tylerjereddy:
Well, the debug code clearly shows `C:\.pythranrc` file with the appropriate contents and with an environment variable set per the docs: https://pythran.readthedocs.io/en/latest/MANUAL.html#customizing-your-pythranrc

Once again, everything looks fine, but pythran is refusing to use `clang-cl.exe` despite its clear presence on the operating system/image. I'm not sure if the issue is with pythran or some other strange component, but we know that this happens on both Azure and Appveyor, so it is not just Appveyor.

At this point, I'll likely proceed with a pythran-diabled windows build for rc3.

rgommers:
Makes sense. I'm fairly sure it's a build system issue, so not just Pythran but an interaction where multiple libraries are all changing how extensions are being built. This should all disappear once we drop `distutils`.

Issue #164, opened 2022-02-04
MAINT: two kinds of appveyor errors


tylerjereddy:
Two different errors are currently preventing success for `1.8.0` final wheel builds on Appveyor (other services are all green already).

This one I haven't seen before, maybe related to connectivity:

```
cd tmp_test
If (Test-Path .\junit-results.xml) {
  (new-object net.webclient).UploadFile(
    "https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)",
    (Resolve-Path .\junit-results.xml)
  )
}
$LastExitCode = 0
Exception calling "UploadFile" with "2" argument(s): "The operation has timed out"
At line:3 char:54
+ ... tps://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)",
+                                                   ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException
 

```

This one we have seen plenty of times, but seems more common in this round of builds:

```
_____________________________ test_quad_vec_pool ______________________________
[gw0] win32 -- Python 3.10.0 C:\Python310-x64\python.exe
C:\Python310-x64\lib\site-packages\scipy\integrate\tests\test__quad_vec.py:108: in test_quad_vec_pool
    res, err = quad_vec(f, -np.inf, np.inf, norm='max', epsabs=1e-4, workers=4)
        f          = <function _lorenzian at 0x00000276BE03D5A0>
C:\Python310-x64\lib\site-packages\scipy\integrate\_quad_vec.py:264: in quad_vec
    res = quad_vec(f2, -1, 1, **kwargs)
        a          = -inf
        args       = ()
        b          = inf
        cache_size = 100000000.0
        epsabs     = 0.0001
        epsrel     = 1e-08
        f          = <function _lorenzian at 0x00000276BE03D5A0>
        f2         = <scipy.integrate._quad_vec.DoubleInfiniteFunc object at 0x00000276D0E3D4B0>
        full_output = False
        kwargs     = {'cache_size': 100000000.0, 'epsabs': 0.0001, 'epsrel': 1e-08, 'full_output': False, ...}
        limit      = 10000
        norm       = 'max'
        points     = None
        quadrature = None
        sgn        = 1
        workers    = 4
[...]

PR #165, opened 2022-02-05
MAINT: disable test uploads

tylerjereddy:
Related to gh-164

* uploading the XML test suite result data to
the Appveyor service has failed for me six
times in a row now, and I need to get this release
out, so this feature branch attempts to disable the
part of the control flow that uploads the XML
data

* the tests themselves should still run, we just
won't see the pretty metadata in the Appveyor
UI for now on the wheels release branch

* `test_quad_vec_pool` may still fail stochastically
b/c of multiprocessing permissions issues on Windows,
but this I can at least restart the failing jobs for,
while the XML upload has a 0 % success rate on 6 consecutive
tries for all matrix entries

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -212,14 +212,14 @@ after_test:
   - python verify_init.py scipy\dist
   - cd tmp_test
   # Upload test results to Appveyor
-  - ps: |
-      If (Test-Path .\junit-results.xml) {
-        (new-object net.webclient).UploadFile(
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Ok, the fix seems to be working for the first Appveyor matrix entry so I'm going to  go ahead and self-merge here to keep the release process moving/rebuild wheels.

tylerjereddy:
This didn't fix the problem unfortunately, it seems it is the upload to the anaconda staging area that fails. I'll try  purging some files from the staging area and reverting this with a push..

PR #169, opened 2022-06-24
MAINT: setuptools pin

### LINKED ISSUE #168: BUG: MacOS/Linux wheels are misnamed
The wheel names start with `SciPy-` instead of `scipy`: https://anaconda.org/multibuild-wheels-staging/scipy/files

I observed identical behavior locally with `paver release` until I pinned `setuptools < 60.0.0`.

Windows just got setuptools pinned for other reasons, I'll try to see if I have an option to pin for the Linux/Mac builds as well (on the release wheels branch for now...). I'll probably leave this open as a reminder for the 1.9.0 release if that doesn't use `cibuildwheel`.

tylerjereddy:
Fixes #168 

I just got burned by this for `1.9.0rc1`. I really need to stop doing backport-only fixes, even when I'm in a rush, except when absolutely necessary.

Anyway, let me confirm that this exact patch fixes the issue on `v1.9.x` before merging here (I pushed back first to prioritize current wheel builds).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -151,7 +151,7 @@ before_install:
           CONTAINER=wheels;
           UPLOAD_ARGS="--no-update-index";
       fi
-    - BUILD_DEPENDS="wheel oldest-supported-numpy $CYTHON_BUILD_DEP $PYTHRAN_BUILD_DEP $PYBIND11_BUILD_DEP"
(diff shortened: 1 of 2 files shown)


tylerjereddy:
wheel names seem fixed on the backport so merging the forward port now

PR #171, opened 2022-06-28
WIP: switch to Meson

rgommers:


rgommers:
Some of the issues are because we have a 6 years old commit pinned in the `scipy` git submodule here. It's remains very hard to understand, why do we even use `scipy` as a submodule rather than just clone it ....

matthew-brett:
Ralf - you found the `BUILD_COMMIT` setting?  That's the thing that should set the scipy commit...

rgommers:
> Ralf - you found the `BUILD_COMMIT` setting? That's the thing that should set the scipy commit...

Thanks, that helps. Still useful to update the submodule I think, will keep that commit.

rgommers:
These warnings are back:
```
C:\Users\appveyor\AppData\Local\Temp\1\build-env-0egqm70c\lib\site-packages\numpy\core\include/numpy/npy_common.h:185:20: warning: 'int _fseeki64(FILE*, long long int, int)' redeclared without dllimport attribute after being referenced with dll linkage
```
but that's only because we are building against older `numpy` versions. Should be harmless.

rgommers:
We're building on Linux now, currently at:
```
Run-time dependency openblas found: NO (tried pkgconfig and cmake)
```
This should be dealt with by setting `PKG_CONFIG_PATH` as discussed in https://github.com/MacPython/openblas-libs/issues/74

The macOS build is at a Bash problem:
```
+ /Users/runner/work/1/s/venv/bin/python -m build --wheel
ERROR Source /Users/runner/work/1/s does not appear to be a Python project: no pyproject.toml or setup.py
```
That's due to this line in `config.sh`, not sure how to pass `repo_dir` along:
```bash
wrap_wheel_builder pyproject_wheel_cmd $@
```
(or maybe a conditional `cd scipy` should do it)

rgommers:
Getting somewhere with the Windows build, it builds the wheel but failed on:
```
python ..\check_installed_package.py
  File "C:\projects\scipy-wheels\check_installed_package.py", line 79, in <module>
    main()
  File "C:\projects\scipy-wheels\check_installed_package.py", line 61, in main
    with io.open(license_txt, 'r', encoding='utf-8') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Python310-x64\\lib\\site-packages\\scipy\\LICENSE.txt'
Command exited with code 1
```
and that file is indeed missing. Need to fix in SciPy `main` first.

rgommers:
> We're building on Linux now, currently at:
> 
> ```
> Run-time dependency openblas found: NO (tried pkgconfig and cmake)
> ```
> 
> This should be dealt with by setting `PKG_CONFIG_PATH` as discussed in [MacPython/openblas-libs#74](https://github.com/MacPython/openblas-libs/issues/74)

The issue is that for SciPy we're stuck on OpenBLAS `0.3.18` (see `tools/openblas_support.py` in the scipy repo), and that version doesn't have pkgconfig files. 0.3.20 does (see issue linked above), but we can't use it because it has bugs that need resolving.

I'll note that the `copy_libs` function in `config.sh` does:
```bash
    basedir=$($PYTHON_EXE scipy/tools/openblas_support.py)
    $use_sudo cp -r $basedir/lib/* $BUILD_PREFIX/lib
    $use_sudo cp $basedir/include/* $BUILD_PREFIX/include
```
so things will work fine with openblas `0.3.18` if we improve the fallback paths in Meson itself. That's a more useful way to spend time than trying to patch `0.3.18` with pkgconfig files in this PR.

rgommers:
The last commit will get the Windows build past the license check failure once `meson-python` 0.7.0 is released.

h-vetinari:
> 0.3.20 [...] has bugs that need resolving.

I tried to look for those in the scipy tracker but did not find any - do you have a link handy?

rgommers:
> I tried to look for those in the scipy tracker but did not find any - do you have a link handy?

- https://github.com/numpy/numpy/issues/21950
- https://github.com/MacPython/openblas-libs/issues/83
- https://github.com/xianyi/OpenBLAS/issues/3659

Looks like the fix is in (but I have not verified that myself). There may be more issues. I do know that numpy and scipy both pass their tests with openblas 0.3.18, so probably that one fix is all we need.

h-vetinari:
> Looks like the fix is in (but I have not verified that myself). There may be more issues. I do know that numpy and scipy both pass their tests with openblas 0.3.18, so probably that one fix is all we need.

If I understood Matthew [correctly](https://github.com/xianyi/OpenBLAS/issues/3659#issuecomment-1164802456) the fix was (or needs to be) on scipy-wheel side, not openblas. As another data point, the test suites of numpy & scipy have been running fine with openblas 0.3.20 in conda-forge.

matthew-brett:
@h-vetinari - yes - the fix is on our side.  I'll merge the relevant PR now, it's been a while.

matthew-brett:
Fix merged from https://github.com/MacPython/openblas-libs/pull/84

rgommers:
Superceded by https://github.com/scipy/scipy/pull/16842, closing.

PR #172, opened 2022-07-27
CI: try to fix nightly versioning, now is `1.10.0.dev0.0.<commit-hash>`

rgommers:
The `.0` should be the commit count, determined in SciPy's `tools/version_utils.py`. Bug report: https://github.com/scipy/scipy/issues/16702


rgommers:
Hmm, this worked fine on TravisCI:
```
$ git -C $REPO_DIR rev-list HEAD v1.8.0 --count
28451
```
So it's a little unclear with the commit count is `0`.

We had about a 5 month gap with broken CI it looks like (or someone cleaned up), so it happened somewhere in that time frame:

image

rgommers:
Wheel name in the first finishing TravisCI job: `scipy-1.10.0.dev0+1555.eecd600-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl`. The `dev0+1555` is what I wanted to see, so in it goes.

tylerjereddy:
thanks, yeah I may have done some purging at some point, it does fill up; we also ran out of credits for a bit and I emailed to get them refilled which they did

PR #174, opened 2022-08-20
DEBUG: Python 3.11 azure testing

tylerjereddy:
* `[skip appveyor] [skip travis]` and temporarily disable most of the Azure CI matrix except for a new `Python 3.11` wheel build job I'm testing
* the PR is against `v1.9.x` because we're planning to switch to `cibuildwheel` for i.e., `1.10`, but if changes on `master` prove necessary for backporting anyway I can just retarget this; for now, I'm just curious to see where this stands with `multibuild`..


tylerjereddy:
That seems to have worked for Azure Linux AMD64 Python 3.11:
- `Fixed-up wheel written to /tmp/tmp.RZwDmJhp9b/scipy-1.9.0-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl` and during build stage I see `/opt/_internal/cpython-3.11.0rc1/lib/python3.11/site-packages`
- `pytest` may be using something a bit older, maybe ok though-- `Python 3.11.0b3`

I'll try to probe the other two CI services next..

tylerjereddy:
The Travis CI Linux ARM64 job also seems reasonable with `3.11` browsing the logs:
- build stage has `/opt/_internal/cpython-3.11.0rc1/lib/python3.11/site-packages` in use and final wheel name `scipy-1.9.0-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl`
- test stage has `Python 3.11.0b3` used by `pytest`, which I suppose is ok

Assuming we're not worried about any of the above, I think I still have to:
- check Appveyor for 64-bit Python 3.11
- check Azure for Python 3.11 with MacOS wheel flavors

charris:
> check Azure for Python 3.11 with MacOS wheel flavors

If you are using multibuld, you need to use Python 3.10 for part of it. Something like
```
          AZURE_PYTHON_VERSION: "3.10"  # azure doesn't have Python 3.11
          MB_PYTHON_VERSION: "3.11"
          MB_ML_VER: "2014"
          DOCKER_TEST_IMAGE: "multibuild/focal_{PLAT}"
```

rgommers:
Build using `rc1` is a must-have indeed; I agree testing with `b3` should be fine although it's slightly puzzling why the same version isn't used.

tylerjereddy:
For MacOS I may need to try `multibuild` update to fix the error, but even then I think I may need to update beyond that upstream to get past the beta version of `3.11`, since I think the last update was: https://github.com/multi-build/multibuild/pull/477/files

rgommers:
Ah fun. Okay, if it's too much trouble then maybe we should leave this for `1.9.2` and `cibuildwheel`. Hard to judge with these things, you never know how much trial and error is needed here.

tylerjereddy:
Looks like the MacOS wheels were built with Python 3.11b5, but the test stage fails because a 3.10 interpreter was used. I'll chip away at it a bit.

tylerjereddy:
NumPy has `setup_test_venv` that we don't have for the testing stage on Azure. I'm going to try to reuse the venv that `multibuild` uses for building the wheel, after purging all packages from that venv so that we get a clean slate for older NumPy and clean install of SciPy wheel just built. That's probably not ideal, but if it works I probably won't complain too much with cibuildwheel on the near horizon.

tylerjereddy:
Alright, this is no fun to debug (as usual) and I haven't even started on Windows, so I'll move on to more interesting stuff for tonight, but I'll leave this open in case I/we get energy to come back to it before `1.9.1`.

PR #176, opened 2022-08-28
Stop cron jobs running on master

rgommers:
We're switching over to cibuildwheel in the main SciPy repo,
so no need for these jobs anymore. The nightlies will be uploaded
from a weekly scheduled run of `cibuildwheel`.

Appveyor doesn't run cron jobs, and TravisCI cron jobs are not defined
in the yml file but in its web UI, so this is all there is to do here.

[ci skip]

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,12 +1,3 @@
-schedules:
-- cron: "27 3 * * 0"
-  # 3:27am UTC every Sunday
-  displayName: Weekly build
-  branches:
-    include:
-    - master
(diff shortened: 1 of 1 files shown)


rgommers:
In case it needs re-adding at some point, for TravisCI the setting is at https://app.travis-ci.com/github/MacPython/scipy-wheels/settings:

image

tylerjereddy (approved):
LGTM, the Azure jobs are already compiling SciPy so there can't be a yml syntax issue at this point.

